PR #27, opened 2016-01-25
Update INSTALL.md

sethbergman:
Spelling correction


dmathieu:
That's not a typo. It's actually what is being displayed. It shouldn't be changed either, as gunzip is the tool being used.
https://github.com/heroku/stack-images/blob/d30a204e40b11742fc872b3006a5ea93e0e80d49/bin/install-stack#L18

PR #36, opened 2016-07-13
Remove workaround for ImageTragick.

ojacobson:
The upstream debs now contain their own feature manifest, so we no longer need to provide our own to protect our customers.


dmathieu:
Nice!

PR #42, opened 2016-09-29
Generate a tar/gzip'ed stack image as well

lstoll:
There are some situations where not having to loop mount a whole image is advantageous, so also generate a straight up tar & gzipped version of the images content before unmounting the created image.
### TODO
- [ ] Use something better than md5 (sha256?)

### DIFF
--- a/bin/capture-stack
+++ b/bin/capture-stack
@@ -73,8 +73,7 @@ EOF
   ) | indent
 
   display Creating tgz image and MD5
-  tar -zcpf $TGZ -C $IMG_MNT ./
-  md5sum $TGZ | cut -d ' ' -f 1 | tr -d '\n' | tee $TGZ.md5 | indent
+  tar -zcp -C $IMG_MNT ./ | tee $TGZ | md5sum | cut -d ' ' -f 1 | tr -d '\n' | tee $TGZ.md5 | indent
 
   display Unmounting image
   umount $IMG_MNT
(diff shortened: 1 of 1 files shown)


tt:
I would really prefer not having to produce a number of different formats.

Building the stack image is already slow (especially because it runs inside a virtual machine with terrible I/O performance) and we sometimes need to do it fast.

I'm happy to consider using another format altogether.

ojacobson:
We could shift to making a tarball the primary artifact, rather than a loopback filesystem. Runtimes can always restore the tarball to a fresh filesystem during installation… Or if we switch to aufs or something, any random directory can be a filesystem root.

rsampaio:
I believe one of the benefits of loopback format in the the way we use the image in cedar is that we can mount the image as read-only straight out of s3 and with tar we have to move to a union filesystem like AUFS or OverlayFS to have this protection from the filesystem level and still there is a memory and performance impact involved, specially for 250 dynos in a single instance also moving the tarball restoration to runtime setup I think isn't a good idea because it would increase the time for an instance to boot (already ~7 min).

Despite the issues I listed here I still think there are benefits of a tar image including the ability to use *any* filesystem (zfs perhaps?) and not having to worry about the number of loopback devices would be good but I can't really see a world where we don't have to generate more than one format so if space or s3 cost is not an issue I would +1 on that.

ojacobson:
@rsampaio Excellent reply. Thank you.

This sounds like a really good justification for an image builder.

danp:
(as part of any change would ❤️  to see a move to something better than md5)

tt:
I'm closing this as we've changed how we're building stack images slightly and will need to rework the approach.

PR #47, opened 2016-11-30
Separate heroku-16 and heroku-dev-16 images

dzuelke:
For now. There are a few packages we still may want to adjust, but this is the biggest chunk so far.

The regular runtime image now has significant space savings over the `-dev` one:

```
After this operation, 408 MB of additional disk space will be used.
```

### DIFF
--- a/heroku-16/bin/heroku-16.sh
+++ b/heroku-16/bin/heroku-16.sh
@@ -85,59 +85,42 @@ PGDG_ACCC4CF8
 apt-get update
 apt-get upgrade -y --force-yes
 apt-get install -y --force-yes \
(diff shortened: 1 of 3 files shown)


heroku-16/bin/heroku-16.sh
```diff
@@ -85,59 +85,42 @@ PGDG_ACCC4CF8
...
-    ruby-dev \
     socat \
     stunnel \
     syslinux \
     tar \
     telnet \
     zip \
-    zlib1g-dev \
-    #
+    libcurl3 \
```
tt: Please sort these alphabetically.

PR #48, opened 2016-12-01
Compute SHA-256 checksum

tt:


dmathieu:
LGTM

RBeaudoin:
Do we still have all of the old MD5's in place still. We need to rollback this change in railgun temporarily and don't want to break the runtime because MD5 is now missing.

ojacobson:
Yes, we still generate and publish the MD5 digests. We won't decommission those until SHA-256 has been in production for quite a while - it's trivial for us to compute and publish them anyways, so there's no rush.

tt:
I already told this but want to mention it again: It's very important that the .md5/.sha256 for each stack image exist on all runtimes to prevent a rerun of [incident 1004](https://status.heroku.com/incidents/1004). All runtimes launched after this pull request shipped will only have the .sha256 file.

PR #63, opened 2017-07-03
Make install-stack create /mnt/stacks earlier

### LINKED ISSUE #62: install-stack step in INSTALL.md fails with "No such file or directory"
Following the cedar-14 steps in INSTALL.md on master, I get:
```
...
$ sudo /vagrant/bin/install-stack /tmp/cedar64-14.4.0.img.gz

-----> Starting install at Mon Jul 3 13:55:15 UTC 2017

-----> Gunzipping image /tmp/cedar64-14.4.0.img.gz
/vagrant/bin/install-stack: line 16: /mnt/stacks/cedar64-14.4.0.img: No such file or directory
```

This is because `/mnt/stacks` doesn't exist in a clean Vagrant instance.

I guess `install-stack` should just do a `mkdir -p /mnt/stacks` before the gzip, just in case it's being run from somewhere that doesn't already have existing images installed.

edmorley:
To ensure the directory exists prior to saving the image within it.

Fixes #62.

### DIFF
--- a/bin/install-stack
+++ b/bin/install-stack
@@ -11,6 +11,8 @@ LOG=/tmp/log/$(basename $0).log
   [ $UID = 0 ] || abort fatal: must be called with sudo
 
   display Starting install at $(date)
+  mkdir -p $MNT
+
   if [[ -f $IMG_GZ ]]; then
     display Gunzipping image $IMG_GZ
     gunzip -c $IMG_GZ > $MNT.img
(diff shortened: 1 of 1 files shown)


tt:
How do you use this script, @edmorley? I'm thinking about removing it as we don't use it internally.

edmorley:
Ah I don't - I was just trying to work out the differences between the image generated using the INSTALL.md process and that output by Docker, after reading #5. By all means remove it if it's unused :-)

On that note -- does heroku-16 use `bin/build-stack` and `bin/capture-stack` like cedar-14 or does it use the Docker image directly? The hardcoding of trusty-specific references in eg `build-stack` makes me think the latter? (Unless there's an internal copy of that file that's been updated for xenial/heroku-16?)

The reason I ask is that I was  going to (a) update `INSTALL.md` with heroku-16 steps, (b) move the cedar-14 specific files to `cedar-14/` (which is either just `cedar-14.sh` or else several of the others too).

PR #66, opened 2017-07-14
Move cedar-14 to its own directory

edmorley:
Three commits:

**1) Move cedar-14 to its own directory**
For parity with Heroku-16.

**2) Add a .dockerignore for cedar-14**
This prevents files that are only used (or generated) by the cedar-14 Vagrant environment from being included in the build context sent to the Docker daemon.

**3) Fix broken push-stack example in README**
Previously the access key/secret were passed as arguments to push-stack rather than setting the environment variables.


edmorley:
Ah one question - are the other scripts in the `bin/` directory cedar-14 specific too? If so I'll move them as well - it just wasn't clear (see https://github.com/heroku/stack-images/pull/63#issuecomment-312679143).

tt:
Everything in the root bin directory is only for Cedar-14 so I think my preference would be moving that to a cedar-14 directory and keeping a bin directory inside each stack for consistency (until we can retire Cedar-14).

edmorley:
I agree - I'll update the PR.

Would you like me to move the `tmp/log/` directory from the repo root, to within the `cedar-14/` directory as well?

tt:
Please do and probably also move the .gitignore file and the Vagrantfile which are only relevant for Cedar-14.

edmorley:
PR updated, however the Travis run didn't trigger.

I spoke to Travis support and they say there is a bug with repos using the new GitHub code owners feature that stops the jobs being scheduled for PRs (they still seem to work for branches, so a run will be triggered once this merges). For now, here's a passing Travis run for this from my fork:
https://travis-ci.org/edmorley/stack-images/builds/254959326

I've moved `tmp/log/*` but left the `Vagrantfile` where it is, since:
* Vagrant hardcodes paths in the metadata files in `.vagrant/` so things would break for any existing VMs.
* Both the cedar-14 scripts and the new `docker-build.sh` script need `stack-helpers.sh`, and moving the Vagrantfile would change the Vagrant root, making the root `bin/` directory inaccessible from the cedar-14 scripts.
...however I've added a comment to the file to make it clearer that it's used by cedar-14 only. 

Sometime soon I'll also open a PR to update INSTALL.md with Heroku-16 build instructions.

edmorley:
Note: After this merges, users that have an existing Vagrant instance must run `vagrant reload` (which re-reads Vagrantfile and reboots the VM), to pick up the updated shared folder location.

tt:
Unless you disagree strongly, I would still suggest moving the .gitignore file and the Vagrantfile (especially since the change already breaks Vagrant machines slightly). Internally, we dispose the Vagrant machines after each stack image build and I think most customers have picked up our Docker images by now.

edmorley:
Makes sense - done. I also added two more commits for some other cleanup - see the multi-line commit messages for more details.

New Travis run (from my fork due to Travis/Github bug mentioned above), here:
https://travis-ci.org/edmorley/stack-images/builds/255754026

In addition, I tested the Vagrant side locally by following the revised INSTALL.md instructions (I used fake credentials so the push-stack step failed with a permissions error, but everything worked up to that point). I then ran `./bin/docker-build.sh cedar-14` from the host to verify that the `.dockerignore` correctly excluded the generated `tmp/log/` files.

tt:
GitHub appears to be broken:

    % git ls-remote git://github.com/heroku/stack-images.git | grep 'refs/pull/66/'
    6607a5789b1d33097d8c30be8106f92ba1304b18        refs/pull/66/head
    86c4b69024f9d80107af53cee14442f809958244        refs/pull/66/merge

However, the current HEAD commit of this branch is db5d085a3711ce7177f768cc865b05a46bba2fdb. 

I contacted their support. Please don't make any changes to this pull request until I hear back.

edmorley:
Huh that's strange - thank you for reporting with them, I'll leave the PR as is.

edmorley:
> I contacted their support. Please don't make any changes to this pull request until I hear back.

@tt, did you hear anything back?

edmorley:
Travis notified me that the GitHub API issues around repos that have the code-owners feature enabled have now been fixed, so I've force pushed the PR to trigger a new Travis run - since GitHub have now had 18 days to look into the revision deviation issue (it may even be due to the same bug). 

The HEAD revision now matches that returned by:
`git ls-remote git://github.com/heroku/stack-images.git | grep 'refs/pull/66/'`

tt:
I heard back that they'd look into it but nothing since. I'm merging this and hopefully it was a one-off on their end.

PR #70, opened 2017-08-14
add missing tzdata package

dzuelke:
no longer included in Ubuntu base images

https://bugs.launchpad.net/cloud-images/+bug/1682305


edmorley:
I wonder if the Ruby buildpack should run its tests inside the `heroku/heroku:16-build` Docker image, like Python buildpack does? Alternatively, testing of the top N buildpacks could be added to this repo's Travis run? (Or both)

tt:
The Ubuntu 14.04 Docker image changed too so if we ever revisit Cedar-14's stack image build process, we could end up affecting it too. I updated the branch to also include it there.

tt:
@ojacobson:

> For a majority of use cases, locale and timezone data takes up extra space without being used.

([This comment](https://github.com/docker-library/official-images/issues/2863#issuecomment-294607003).)

tt:
@edmorley, we should certainly improve our testing to include an even broader array of test apps. None of our internal components or trimmed down test candidates were affected by the missing package.

We'll also improve our roll-out checklist so a diff of packages is generated prior to the roll-out itself. It's currently the last task (to prevent updating documentation before it applies, I believe).

PR #71, opened 2017-08-16
Fix libicu52 regression

dzuelke:


dzuelke:
Bunch of stuff got removed yesterday. Checking for more missing packages that may impact customers.

PR #73, opened 2017-08-16
Export lists of installed packages and validate on Travis

edmorley:
The docker-build.sh script now exports the list of all packages installed in the final images to the working directory. These package lists are checked-in, making any deviations both easy to spot locally plus allowing for their validation on Travis.

As a one-off for this PR, I based `cedar-14/installed-packages.txt` not on the output from `docker-build.sh` but instead on the output from `dpkg-query` run on a cedar-14 dyno (that is using the older image after the rollback, that has the correct packages installed). 

As such, the cedar-14 part of the Travis run should hopefully fail and in the diff show the regression that is about to be fixed in #71, demonstrating the check works.


edmorley:
There are actually quite a lot of packages (in addition to the known libicu52) missing from a build of cedar-14 here, compared to the older cedar-14 dyno image:
https://travis-ci.org/heroku/stack-images/jobs/265101108

This is puzzling, since the cedar-14 docker image uses `ubuntu-debootstrap:14.04` which hasn't been updated in over a year:
https://hub.docker.com/r/library/ubuntu-debootstrap/tags/

Has the way the production cedar-14 image is built been changed internally, and so it no longer reflects the Dockerfile here?

tt:
> Has the way the production cedar-14 image is built been changed internally, and so it no longer reflects the Dockerfile here?

The Cedar-14 stack image is still built using tools available in this repository. [It executes debootstrap](https://github.com/heroku/stack-images/blob/6a4ad47f68d67e3c6a21702010138d60f32e4dfa/cedar-14/bin/build-stack#L33) on each build. The result of that doesn't appear to be any different from the Docker image, though.

I think the previous existence of FUSE is particularly suspicious. I'll try to dig into that with `apt-cache rdepends`. I'm wondering how often package relationship changes upstream or if maybe this was an unintended side effect of 1c1a386d1fce09eafff24a54cee9ea7bee72a9f0.

tt:
Overall, this change looks good.

We held a retrospective on a stack image roll-out yesterday and a snapshot of installed packages is something we've wanted to do for a while.

I'm wondering if we want to keep an entire graph in the source tree (but limiting to roots in the diff). That might be useful for later understanding how things change.

.travis.yml
```diff
@@ -8,3 +8,9 @@ env:
   - STACK=cedar-14
 script:
   - bin/docker-build.sh $STACK
+  - |
```
ojacobson: I have a mild preference that this be written as a shell script in a file, rather than inline in `.travis.yml`, as it makes the travis file a bit easier to read if the task has a name rather than just a bolus of code.

edmorley:
> The Cedar-14 stack image is still built using tools available in this repository. It executes debootstrap on each build. The result of that doesn't appear to be any different from the Docker image, though.

Running `dpkg-query --show --showformat='${Package}\n'` on the cedar-14 docker image built from master and comparing it to the same run against the image built via `build-stack` inside the Vagrant environment, shows that the package list has one difference - `inetutils-ping` is present in the Vagrant generated image, but not the Docker file.

This appears to be due to the `ubuntu-debootstrap:14.04` Docker image passing `--include` to debootstrap, which `build-image` does not. Compare:

https://github.com/tianon/docker-brew-ubuntu-debootstrap/blob/61e5df54c69a8d3943546e26b15d3fff80254e45/14.04/build-command.txt#L1
https://github.com/heroku/stack-images/blob/6b7229c65b4286c018c7ba36ba4ef6fa0bfb2f55/cedar-14/bin/build-stack#L33

That said, it's an extra package and not the cause of the missing packages, so orthogonal :-)

To figure out the package dependency changes, I ran the following against both a cedar-14 dyno and the local docker image built from master:
`dpkg-query --show --showformat='${Package}\n  Version: ${Version}\n  Depends: ${Depends}\n  Recommends: ${Recommends}\n  Suggests: ${Suggests}\n\n'`

Output:
* [From dyno](https://github.com/heroku/stack-images/files/1231849/packages.txt)
* [From local Docker image](https://github.com/heroku/stack-images/files/1231853/packages.txt)

Diffing these and focusing on packages that are present in both, but that have different "depends" or "recommends" ("suggests" aren't installed by default so are irrelevant here), found the following differences:
* openjdk-7-jre: Recommends no longer contains `libgnome2-0, libgnomevfs2-0, libgconf2-4`
* libgraphviz-dev: Depends now contains `libgvc6-plugins-gtk` when it didn't before

In the Docker image locally, running `apt-get update && apt-get install libgnome2-0 libgnomevfs2-0 libgconf2-4` shows:
```
  acl at-spi2-core colord dbus dbus-x11 dconf-gsettings-backend dconf-service
  desktop-file-utils dmsetup dosfstools eject fuse gconf-service
  gconf-service-backend gconf2 gconf2-common gdisk groff-base gvfs gvfs-common
  gvfs-daemons gvfs-libs libapparmor1 libatasmart4 libatk-bridge2.0-0
  libatspi2.0-0 libavahi-glib1 libbonobo2-0 libbonobo2-common libcanberra0
  libcolord1 libcolorhug1 libdbus-glib-1-2 libdconf1 libdevmapper1.02.1
  libfontenc1 libfuse2 libgconf-2-4 libgconf2-4 libgnome2-0 libgnome2-bin
  libgnome2-common libgnomevfs2-0 libgnomevfs2-common libgphoto2-6
  libgphoto2-l10n libgphoto2-port10 libgtk-3-0 libgtk-3-bin libgtk-3-common
  libgudev-1.0-0 libgusb2 libicu52 libidl-common libidl0 libieee1284-3
  liborbit-2-0 liborbit2 libpam-systemd libparted0debian1 libpolkit-agent-1-0
  libpolkit-backend-1-0 libpolkit-gobject-1-0 libsane libsane-common
  libsecret-1-0 libsecret-common libsystemd-daemon0 libsystemd-login0 libtdb1
  libudisks2-0 libusb-1.0-0 libv4l-0 libv4lconvert0 libvorbisfile3
  libwayland-client0 libwayland-cursor0 libxaw7 libxcb-shape0 libxft2
  libxkbcommon0 libxmu6 libxv1 libxxf86dga1 ntfs-3g parted policykit-1
  policykit-1-gnome psmisc sound-theme-freedesktop systemd-services
  systemd-shim udisks2 x11-utils xkb-data
```

After that, re-running the `dpkg-query` command from above gives:
[Output](https://github.com/heroku/stack-images/files/1232014/packages-new.txt)

This is much closer to the dyno output, but there is now another package that can be seen to have changed its dependencies:
* systemd-services: It previously had a depends on `systemd | systemd-shim` which has now changed to just `systemd-shim`

Installing `systemd` gives us `libcryptsetup4 libsystemd-journal0 systemd`, which are the last of the missing packages.

Running `apt-cache policy systemd-services openjdk-7-jre` gives:
```
systemd-services:
  Installed: (none)
  Candidate: 204-5ubuntu20.24
  Version table:
     204-5ubuntu20.24 0
        500 http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
     204-5ubuntu20 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
openjdk-7-jre:
  Installed: 7u131-2.6.9-0ubuntu0.14.04.2
  Candidate: 7u131-2.6.9-0ubuntu0.14.04.2
  Version table:
 *** 7u131-2.6.9-0ubuntu0.14.04.2 0
        500 http://archive.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
        500 http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        100 /var/lib/dpkg/status
     7u51-2.4.6-1ubuntu4 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
```

Both of these are in `main` rather than `universe`, so I believe the dependency changes were unrelated to #61.

It seems surprising that the dependencies would be allowed to change in a stable LTS release for a `main` package, but perhaps that's permitted for "recommends" more so than "depends"? For Heroku-18 maybe it's worth using `--no-install-recommends` with apt-get and specifying those packages explicitly to increase the chance of a stable package list?

.travis.yml
edmorley: I agree -- originally I was hoping it would be more of a one-liner but with trying to handle both the "dirty working directory" and "new untracked files" case that isn't possible with Git at present. Perhaps this can go in `bin/runtests.sh` and in the future additional tests be added there (though admittedly failing tests if the working directory is dirty does seem weird locally..).

I'll hold off updating this PR until other PRs add whatever missing packages are deemed required (I'm presuming the whole of `libgnome2-0`, `libgnomevfs2-0`, `libgconf2-4`, `systemd` plus deps might be unnecessary?) .

tt: I would actually prefer that it stayed in the Travis file. It's only used there and I'd like to get rid of the top-level bin directory.

tt: @edmorley, unfortunately, some of those packages are used by [heroku-buildpack-google-chrome](https://github.com/heroku/heroku-buildpack-google-chrome) and while it's unlikely that it's being used at runtime, it's an unnecessary risk.

tt:
@edmorley, can you try rebasing this to trigger a new build?

edmorley:
@tt done. I left the old package list (from existing dynos) in, so the remaining missing packages can be seen.
Once you're fine with what's there, I can change it to be what is generated locally.

tt:
The new list looks much better. Please update the checked in version and I'll merge this pull request.

edmorley:
Done

tt:
Thank you very much for this, @edmorley.

edmorley:
You're welcome :-)

PR #74, opened 2017-08-22
Depend explicitly on libgconf-2-4 package

tt:
[heroku-buildpack-google-chrome](https://github.com/heroku/heroku-buildpack-google-chrome) expects this package:

https://github.com/heroku/heroku-buildpack-google-chrome/blob/03127f6d1c8bc7f8793a01f943ca448e4f91ec52/bin/compile#L63-L73

PR #75, opened 2017-08-22
Depend explicitly on libgtk-3-0 package

tt:
[heroku-buildpack-google-chrome](https://github.com/heroku/heroku-buildpack-google-chrome) expects this package:

https://github.com/heroku/heroku-buildpack-google-chrome/blob/03127f6d1c8bc7f8793a01f943ca448e4f91ec52/bin/compile#L63-L73


dmathieu:
What size is this adding to the stack image? Could the buildpack be installing that itself? 
I'm only commenting once, but the same question applies to #74.

tt:
The package used to be in the stack image so we're not adding it. We're making the dependency explicit so it remains installed.

PR #77, opened 2017-08-23
Depend explicitly on psmisc package

tt:


tt:
(This should fail but want to check the new Travis script in action.)

ojacobson:
Failure diff looks good!

PR #85, opened 2017-10-30
Some MacOS support fixes.

ojacobson:
* No `tac` on OS X, even in Homebrew. However, tail fills the same job.
* Don't use gnu sed's `-u` unbuffered mode, it's not available on MacOS under BSD sed.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -34,6 +34,9 @@ if [[ "$STACK" != "cedar-14" ]]; then
     write_package_list "$BUILD_IMAGE_TAG" "$BUILD_DOCKERFILE_DIR"
 fi
 
+# MacOS protip: if this step fails, run `brew install coreutils` and follow the
+# directions in the output to update your PATH and MANPATH. `tac` is part of
+# that package.
 display "Size breakdown..."
 docker images --format "table {{.Repository}}:{{.Tag}}\t{{.Size}}" \
     | grep -E "(ubuntu|heroku)" | tac | indent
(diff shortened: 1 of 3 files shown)


docker-build.sh
```diff
@@ -34,6 +34,15 @@ if [[ "$STACK" != "cedar-14" ]]; then
     write_package_list "$BUILD_IMAGE_TAG" "$BUILD_DOCKERFILE_DIR"
 fi
 
+case "$(uname)"
+Darwin)
+  TAC=tail -r
+  ;;
+*)
+  TAC=tac
```
tt: Is there any reason to not use `tail -r` consistently?

edmorley:
> I don't understand why you need to drop -u. You can't run the Ubuntu-based installation script directly on macOS anyway.

The docker-build.sh script is being run from the host to regenerate the installed packages list, and it uses `indent()`. I presume the other instances were modified for consistency.

(And sorry for the non-OSX compatible `tac` usage; it always seems hit and miss with OS X and cross-platform scripting 😕 )

tt:
That's a good point. I probably relied on Travis to run that script.

tt:
@ojacobson, please still use `tail -r` on all platforms if possible.

ojacobson:
> @ojacobson, please still use tail -r on all platforms if possible.

Hah, I wish, but `tail -r` is a BSD-ism not available under GNU Coreutils. (I checked, I couldn't find tac in Homebrew, either.)

```
~$ tail -r example.txt
tail: invalid option -- 'r'
Try 'tail --help' for more information.
```

PR #89, opened 2018-01-02
Upgrade postgresql client to 10

dmathieu:
cc @keiko713 @heroku/dod-core

PR #90, opened 2018-01-03
Add GNU gettext

codingjoe:
[gettext][gettext] is commonly used in various frameworks and
programming languages to provide support for i18n. Such as
* Python/Django
* Ruby/translation
* Python/flask-babel
* ... may more.

I does not make sense to have it available in side the runtime dyno, but during build time, it is helpful to have the package available to compile gettext `po` files (to `mo`).

There are already some buildpacks doing that, but since this is a GNU tool, it makes sense to have it in the main image.

[gettext]: https://www.gnu.org/software/gettext/

### DIFF
--- a/heroku-16-build/bin/heroku-16-build.sh
+++ b/heroku-16-build/bin/heroku-16-build.sh
@@ -9,6 +9,7 @@ apt-get install -y --force-yes \
     autoconf \
     bison \
     build-essential \
+    gettext \
     libacl1-dev \
     libapparmor-dev \
     libapt-pkg-dev \
--- a/heroku-16-build/installed-packages.txt
+++ b/heroku-16-build/installed-packages.txt
@@ -52,7 +52,6 @@ gcc-6-base
 geoip-bin
 geoip-database
 gettext
-libasprintf-dev
 ghostscript
 gir1.2-freedesktop
 gir1.2-gdkpixbuf-2.0
--- a/heroku-16-build/installed-packages.txt
+++ b/heroku-16-build/installed-packages.txt
@@ -49,9 +49,10 @@ gcc
 gcc-5
 gcc-5-base
 gcc-6-base
-gettext
 geoip-bin
 geoip-database
+gettext
+libasprintf-dev
 ghostscript
 gir1.2-freedesktop
 gir1.2-gdkpixbuf-2.0
@@ -97,6 +98,8 @@ libapt-pkg-dev
 libapt-pkg5.0
 libasan2
 libasn1-8-heimdal
+libasprintf-dev
+libasprintf0v5
 libatk1.0-0
 libatk1.0-data
 libatm1
@@ -192,6 +195,8 @@ libgdk-pixbuf2.0-common
 libgdk-pixbuf2.0-dev
 libgeoip-dev
 libgeoip1
+libgettextpo-dev
+libgettextpo0
 libgirepository-1.0-1
 libglib2.0-0
 libglib2.0-bin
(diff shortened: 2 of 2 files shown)


codingjoe:
@tt what needs to be done to get this thing on the road? Anything I can do?

heroku-16-build/bin/heroku-16-build.sh
```diff
@@ -9,6 +9,7 @@ apt-get install -y --force-yes \
     autoconf \
     bison \
     build-essential \
+    gettext \
```
ojacobson: Hi, and thanks for the contribution!

One question. Putting the actual `gettext` package here will install the `.so` files in the `heroku-16-build` image, but not in the `heroku-16` image. The library will be available at build-time only. Is this intended?

codingjoe: Yes, one should compile gettext files only during deploy not during runtime.

dmathieu:
@codingjoe could you give us the size of the package when installed?

codingjoe:
> @codingjoe could you give us the size of the package when installed?
@dmathieu what's the recommended why to measure that? Only package or including dependencies?

edmorley:
The Travis script outputs the image sizes to the log to help with decisions like these :-)

master: 
       heroku/heroku:16         461MB
       heroku/heroku:16-build   852MB

HEAD~2 (this PR's parent until it's rebased, and so doesn't contain #89):
       heroku/heroku:16         460MB
       heroku/heroku:16-build   851MB

This PR:
       heroku/heroku:16         461MB
       heroku/heroku:16-build   858MB

dmathieu:
Including dependencies. Basically, what's the stack image size difference before and after adding this package.

codingjoe:
6MB than, thanks @ojacobson 
How often/When do you release new images?

dmathieu:
@codingjoe we release stack images when required. It just happens that we're releasing one today (without your patch).
We could release again next week though. I'd like to get some else from @heroku/build to approve the PR though.

dmathieu:
ping @dzuelke too.
WDYT? Especially about having it only in -dev?

codingjoe:
> @codingjoe we release stack images when required. It just happens that we're releasing one today (without your patch).
We could release again next week though. I'd like to get some else from @heroku/build to approve the PR though.

sounds cool, always a pleasure to work with you guys 🥇

dzuelke:
Yeah the dev-only part is fine @dmathieu. Really all that's needed is `msgfmt` but we can only install the entire package. I suppose many people still produce `.mo` files locally and check them into version control, but thinking about it, it actually makes sense to do so at build time.

codingjoe:
> I suppose many people still produce .mo files locally and check them into version control, but thinking about it, it actually makes sense to do so at build time.
My point exactly 👍

codingjoe:
Awesome, docker works. How long does it take until the new image is available during the regular buildpack builds?

dmathieu:
I'll let you know here once I've finished updating the stack image.

dmathieu:
It's out: https://devcenter.heroku.com/changelog-items/1356

codingjoe:
Nice, I do get a `msgfmt: Command not found` but I guess I just need to wait a bit.

dmathieu:
Yes, not all instances have finished updating yet. That'll be done in the next couple of hours.

codingjoe:
I still get `make: msgfmt: Command not found` even after two days. I call `msgfmt` from `post_compile` inside the Python buildpack :/

dmathieu:
@codingjoe we're aware of the issue which is not related to your patch. We're working on it and will let you know once the stack image is shipped again.

dmathieu:
@codingjoe sorry for the delay here. gettext should be available now.

codingjoe:
🎉 I can confirm it works 👍

PR #91, opened 2018-01-12
Add mupdf and ffmpeg for Rails 5.2 ActiveStorage support

hone:
This issue was opened on the ruby buildpack:
https://github.com/heroku/heroku-buildpack-ruby/issues/635

ActiveStorage has native support for PDF previews with the mupdf tool as
well as metadata information with ffmpeg. ActiveSupport shells out to
these binaries directly.

Here are the size changes from adding these packages in:

heroku/heroku:16
base: 461MB
mupdf: 471MB
mupdf+ffmpeg: 645MB

heroku/cedar:14
base: 1.35GB
mupdf: 1.36GB
mupdf+ffmpeg: 1.46GB

This amounts to a 184mb increase in stack image size.


dmathieu:
:-1: this is a huge overhead for something the vast majority of apps don't need.

ojacobson:
I tend to agree with @dmathieu on this. Is there another way we can practically meet the need in heroku/heroku-buildpack-ruby#635, such as a mupdf buildpack?

Longer-term, this is a good signal that we need to be thinking about how to support package installation on Heroku, but we're not there (yet).

edmorley:
> Longer-term, this is a good signal that we need to be thinking about how to support package installation on Heroku, but we're not there (yet).

I haven't tried this particular feature of build manifests yet, but it looks like it's a first class citizen when using it (compared to having to use [heroku-buildpack-apt](https://github.com/heroku/heroku-buildpack-apt) at least):
https://devcenter.heroku.com/articles/heroku-yml-build-manifest#installing-packages-from-apt

dmathieu:
@edmorley I think what Owen talks about is providing a way for buildpacks to install packages.

ojacobson:
@edmorley @dmathieu Yeah, I'd somewhat forgotten that it's okay to talk about the heroku.yml feature now that it's in dev preview, and wanted to avoid accidentally committing the team to anything.

I do still think there's more to do; we're still working out how to (and, really, if we should) converge heroku.yml and the current build and runtime environments.

ojacobson:
I had a conversation with @hone this afternoon to try to reach a decision about this pull request. Some key points that I'm not sure we'd fully considered:

* Users are specifically requesting support for the Rails 5 image and PDF preview features (see eg. https://github.com/heroku/heroku-buildpack-ruby/issues/635). We can take as read that @hone will find a way to support these users and meet that need regardless of whatever decision we make on this pull request.

* ffmpeg, one of the implicit dependencies of this Rails feature, is a whopping 174MB in size. This library must be present for the Rails 5 preview feature to function, so it either needs to be part of the stack image, part of the slug, or downloaded during app boot in some way. Downloading ffmpeg every time a Rails dyno boots, either in the slug or after slug startup, is a fairly substantial boot time penalty, and will impact our COGS by substantially increasing the amount of data we load from S3 on each dyno boot. Placing ffmpeg in the stack image means we only have to download it once, when each runtime instance boots, and that it can be shared between slugs.

* Adding 174MB to the stack Docker images will make them bulkier to work with and slower to download. It'll also affect the stack image rollout process by a bit.

I'm convinced I should change my mind and support this pull request.

titanous:
One thing that hasn't been brought up is security. ffmpeg is notoriously buggy due to its use of memory-unsafe code (aka C), for example, OSS-Fuzz has found many critical vulnerabilities: https://bugs.chromium.org/p/oss-fuzz/issues/list?can=1&q=project-ffmpeg+Type%3DBug-Security

A single vulnerability in ffmpeg or mupdf as currently implemented in Rails 5.2 can result in the complete takeover of the application due to remote code execution via untrusted user input. I would strongly recommend not implementing ffmpeg/mupdf without strong sandboxing.

I recently merged a PDF fuzzer into OSS-Fuzz for mupdf, but a sandbox is still necessary.

dmathieu:
> Rails applications don't give us any handles for detecting "uses ffmpeg" reliably.

In my opinion, this is the issue here. Dependencies should be explicitly declared, and this brings 2 huge dependencies implicitly to everyone.
In order for us to properly support this, rails needs to give us a way to know if an app needs these dependencies.
Then, we can selectively install the package in a slug if it needs it. If there are slug size issues, we could issue exceptions for apps that need this package.

In addition to security, the other very concerning thing in my opinion is [mupdf's license](https://github.com/rails/rails/pull/30667#issuecomment-332276198).
Adding the package to every slug, all our customers would essentially be breaking the license terms, as they'd be using it in a closed-source application.
Bringing this package magically could be more of a curse than a blessing.

ffmpeg is not MIT either.

dmathieu:
Re: security and potential future security issues.
Having it in the stack image would actually be better there, as we'd be able to upgrade for all apps without requiring customer to redeploy.

cji:
I'm also 👎  on including this in the base stack image from a security perspective. I agree 💯 with @titanous on ffmpeg's history (thanks for getting mupdf a fuzzer!). Having file processing libraries like ffmpeg, mupdf, even imagemagick sandboxed would be much more ideal, but is another project. 

We're increasing the attack surface for all applications by including this package in the stack image. While it would allow us to upgrade all users at once with a stack image change, it's an additional 2 packages to monitor for security issues and ensure the stack image is kept up to date.

It also sounds like the licensing issues @dmathieu raises makes this more problematic.

tt:
This pull request was meant to gather data around what packages would be installed and how much space they would take up.

I'm closing it for now and will schedule a meeting to follow up on this.

PR #93, opened 2018-01-25
Build a preview Heroku-18 image.

ojacobson:


heroku-18/bin/heroku-18.sh
```diff
@@ -0,0 +1,144 @@
...
+    stunnel \
+    syslinux \
+    tar \
+    telnet \
+    tzdata \
+    wget \
+    zip \
+
+# install the JDK for certificates, then remove it
+apt-get install -y --no-install-recommends ca-certificates-java openjdk-8-jre-headless
```
tt: It doesn't matter much when we remove it immediately but "openjdk-8-jre-headless" but I think we should use one of the more recent versions (probably whatever is the default for that Ubuntu distribution).

ojacobson:
@dmathieu done!

heroku-18-build/bin/heroku-18-build.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+exec 2>&1
+set -e
+set -x
+
+apt-get update
+apt-get install -y --force-yes \
```
edmorley: What are your thoughts on adding `--no-install-recommends` for all apt-get install invocations (given the clean slate for a new stack) - and then adding back any still-desired packages explicitly? 

Currently there are likely packages being included which are not used, and only serve to increase the image size.

ojacobson: Not a bad idea, and this is a good opportunity to do that since we don't have any compatibility guarantees for this future stack yet. I do worry about the potential impact on people upgrading from Heroku-16 (stealth dependencies being what they are) but we can probably deal with that as it happens.

edmorley: Trying `--no-install-recommends` locally:
* `heroku/heroku:18`: 475MB -> 331MB
* `heroku/heroku:18-build`: 892MB -> 774MB

(Looking at the `installed-packages.txt` diff there are some obvious candidates that would want to be added back straight away - such as `unzip` and `lsb_release` -- though even if lots of packages were added back, listing them explicitly seems safer given #72 and related in the past.)

ojacobson:
Holding until the Ubuntu 18.04 release in April.

PR #96, opened 2018-03-22
Add apt-transport-https package

cji:
Allows us to support using TLS hosted apt repositories in the stack image and any apt buildpacks (https://github.com/heroku/heroku-buildpack-apt/issues/26)

### DIFF
--- a/heroku-16-build/bin/heroku-16-build.sh
+++ b/heroku-16-build/bin/heroku-16-build.sh
@@ -6,7 +6,6 @@ set -x
 
 apt-get update
 apt-get install -y --force-yes \
-    apt-transport-https \
     autoconf \
     bison \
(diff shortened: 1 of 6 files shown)


heroku-16-build/bin/heroku-16-build.sh
```diff
@@ -6,6 +6,7 @@ set -x
 
 apt-get update
 apt-get install -y --force-yes \
+    apt-transport-https \
```
ojacobson: Formally unnecessary, since h-16-build relies on h-16, but harmless to have here.

cji: Ah, missed that in the dockerfile. will fix to make it cleaner/clearer.

PR #97, opened 2018-04-10
Rails 5.2 ActiveStorage Support

hone:
This adds the packages `poppler-utils` and `ffmpeg` to the cedar-14 and heroku-16 stack images.


titanous:
As discussed in #91, there is a major security risk to including these packages without proper sandboxing.

dzuelke:
Who's "mc3man", and why are we using packages from his PPA and not from official vendor repositories?

schneems:
@titanous we talked about it offline with the security team. The general consensus was that for the customers using these features, it's more secure to roll this into the stack image so that customers can get security patches without requiring additional actions. (i.e. if it's put somewhere else, like a 3rd party buildpack, it would require an additional deploy before people are protected). 

Since this is needed to support a key feature in Rails 5.2 people will definitely be using it. Of those that are using it, they'll be safer if they can get security patches without having to take action.

It's also worth nothing that a key difference between this PR and #91 is that we're using poppler for PDF which is GPL instead of mupdf which is AGPL. There was a licensing concern with mupdf, but not with poppler. Support for poppler was added into Rails by Terence to alleviate this concern.

PR #100, opened 2018-04-13
Reverse lines using sed

tt:
This replaces #98 and #99 and emulates the tac tool's behavior.

(The command comes from [this page](http://web.archive.org/web/20090204080114/http://student.northpark.edu/pemente/sed/sed1line.txt).)


tt:
This doesn't suffer from the limitations of numerical sorting and works as intended for the Cedar-14 image too.

PR #104, opened 2018-05-16
Add iproute2 package to heroku-18

jkutner:
This will add the `ip` command, which is already present on cedar-14 and heroku-16. It is required for [heroku-exec](https://devcenter.heroku.com/articles/exec).

### DIFF
--- a/heroku-18-build/installed-packages.txt
+++ b/heroku-18-build/installed-packages.txt
@@ -79,6 +79,7 @@ imagemagick
 imagemagick-6-common
 imagemagick-6.q16
 init-system-helpers
+iproute2
 iputils-tracepath
 krb5-multidev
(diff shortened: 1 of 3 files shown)


jkutner:
@tt done

jkutner:
@dmathieu this will end up blocking heroku-18 from shipping, so you might want to do it yourself if I don't get to it soon.

edmorley:
If it helps, the quick way to ensure all the necessary files are updated (and because local testing is quicker than waiting for Travis :-)), is to follow the steps here:
https://github.com/heroku/stack-images/blob/master/BUILD.md#building-stack-images-locally

After building the images, the script updates the `installed-packages.txt` files locally for you.

PR #110, opened 2018-07-19
Build docker images on travis

dmathieu:
Don't rely on DockerHub anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 - docker
 branches:
   only:
-  - master
+  - automation
 env:
   matrix:
   - STACK=cedar-14
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,4 @@ deploy:
   provider: script
   script: sh bin/publish.sh
   on:
-    branch: automation
+    branch: master
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,14 @@ branches:
   - automation
 env:
   matrix:
-  - STACK=cedar-14
-  - STACK=heroku-16
-  - STACK=heroku-18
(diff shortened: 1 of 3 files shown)


beanieboi:
works for me :)

.travis.yml
```diff
@@ -2,19 +2,29 @@ sudo: required
 dist: trusty
 language: shell
 services:
-  - docker
```
tt: Why are these no longer indented? [It's the default YAML syntax](http://yaml.org/spec/1.2/spec.html#id2759963).

dmathieu: It was regenerated to use Travis' generator

edmorley: Fwiw prettier now supports yaml (https://prettier.io/blog/2018/07/29/1.14.0.html#yaml)... :-)

beanieboi: I run `travis lint` before I commit changes to `.travis.yml` and it does flatten the structure as well

docker-build.sh
```diff
@@ -5,10 +5,10 @@ set -euo pipefail
 cd "$(dirname "${BASH_SOURCE[0]}")"
 . stack-helpers.sh
 
-[ $# -eq 1 ] || abort usage: $(basename "${BASH_SOURCE[0]}") STACK
+[ $# -eq 2 ] || abort usage: $(basename "${BASH_SOURCE[0]}") STACK
```
edmorley: `BUILD.md` needs updating to reflect two args must now always be passed. Or alternatively, `IMAGE_TAG` below could have a fallback default.

dmathieu: Thanks. I'll update that doc.

PR #116, opened 2018-09-18
Add missing runtime packages from heroku-16 to heroku-18

jabrown85:
I went through [this](https://devcenter.heroku.com/articles/stack-packages) and looked for all packages that were in `heroku-16` at runtime and were not in `heroku-18` at runtime.

Reference: heroku/heroku-buildpack-google-chrome/issues/59


jabrown85:
I have a couple questions.

1. When building a stack, how do we determine which packages get added (or left behind)?
2. How does https://devcenter.heroku.com/articles/stack-packages get updated once we publish an update?

tt:
Are these all required for that buildpack?

jabrown85:
@tt I gave up doing it one by one when about 15 of them failed in a row. If there is a particular one you are worried about, I could try removing it.

jabrown85:
We could also move these to the buildpack itself if we rather. I was under the assumption any missing packages were likely an oversight.

ojacobson:
> We could also move these to the buildpack itself if we rather. I was under the assumption any missing packages were likely an oversight.

I believe I gave you that impression. That's my understanding as well. However…

> When building a stack, how do we determine which packages get added (or left behind)?

… this process hasn't been terribly formal. The closest we got was in the design of the Heroku-16 stack, when the build team specifically prioritized shrinking the stack relative to Cedar-14 to make the Docker image more usable for customers.

tt:
@jabrown85, I'm not concerned about any in particular but if we add packages we don't need, we'll also never know if they're safe to remove.

Is it possible to make a simple test script? If so, we can test in a local Docker container where `apt-get install` is much faster than having to rebuild the stack image.

edmorley:
It's worth noting that the difference in package lists between Heroku-16 and Heroku-18 is more of a feature than a bug, given:
https://github.com/heroku/stack-images/pull/93#discussion_r163893594

jabrown85:
It looks like these are not installed when doing `dpkg -x chrome.deb` on `heroku-18`

```
libdjvulibre-text
libdjvulibre21
libglib2.0-data
libilmbase12
libkmod2
libnetpbm10
libopenexr22
libwmf0.2-7 
```

Should we remove these packages from the PR or, given the comment above, should we install all these via chrome buildpack? If this is the only case we are aware of that needs these, I'm fine with that.

jabrown85:
I've removed the packages that we don't currently know if we need to support in `heroku-18`. The packages left are required to run chrome/webkit, imagemagick, wkhtmltopdf and other common scenarios.

merwok:
Nice to see this merged!  I don’t know how Heroku stack versioning works, can I ask what’s the mechanism to get the new version for our apps?

beanieboi:
we are testing them internally first and roll them out after testing, next week is special tho with Dreamforce coming up, we won't change stack images next week but you can expect them to be automatically available beginning October 1st.

merwok:
Would you mind expanding on what «available» means?

I am using old-style config with runtime.txt and Procfile, not app.json or haroku.yml, so I think the only config knob I have is the coarse-grained choice of `Stack: heroku-18` in my dashboard settings.  Are updates such as this one automatically pulled by the next slug build?

dmathieu:
Once we update the stack image in production, we will post a [changelog entry](https://devcenter.heroku.com/changelog). Then, every app running on that stack (heroku-18 in this case) will be on that new version once their app is restarted (via a deployment, or a [daily cycle](https://devcenter.heroku.com/articles/dynos#automatic-dyno-restarts)).

24 hours after we update the stack image, every app on that stack runs the new version.

PR #125, opened 2018-11-01
publish images if we run a tag

beanieboi:
right now we are only building the images on Travis and do not publish them to Docker Hub


dmathieu:
This means we don't push the nightly when pushing to master (or on the weekly cron) anymore. We only push on new releases.

beanieboi:
@dmathieu ok, then we need to change it again. before this change it was the other way around it wouldn't build on new releases. i needed that change to build a new release image

PR #129, opened 2019-01-08
add libmagickcore extra codecs for heroku-18

dzuelke:
this package is required for e.g. SVG manipulation

heroku-16 has it, but not through an explicit dependency, so we're adding that as well


dmathieu:
Before:

```
-----> Size breakdown...
       ubuntu:18.04             86.7MB
       heroku/heroku:18         501MB
       heroku/heroku:18-build   794MB
```

After:
```
-----> Size breakdown...
       ubuntu:18.04             86.7MB
       heroku/heroku:18         507MB
       heroku/heroku:18-build   794MB
```

PR #130, opened 2019-02-12
Add a docker tag in the `heroku/heroku:18.2019-02-12-10.23.22` format.

jabrown85:
This will allow for users greater control over pulling specific images.

Ref: W-5549708

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ branches:
   - /^v\d+(\.\d+)?(\.\d+)?(-\S*)?$/
 env:
   matrix:
-  - STACK=cedar-14 IMAGE_TAG=heroku/cedar:14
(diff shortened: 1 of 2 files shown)


tt:
Why do we want to allow users pulling a potentially vulnerable image?

jabrown85:
@tt it might be more for us internally during stack image updates/rollback. I can't locate the incident that spawned this request at the moment.

jabrown85:
@tt updated to push to the private repo instead.

edmorley:
> it might be more for us internally during stack image updates/rollback. I can't locate the incident that spawned this request at the moment.

I managed to track this down to [PRB-0005311](https://gus.lightning.force.com/lightning/r/SM_Problem__c/a4QB00000004XODMA2/view).

PR #132, opened 2019-02-14
Adding tools for converting docker stack to heroku stack

jabrown85:
Add tooling to convert to heroku stack and ability to publish from build system to internal staging area.


jabrown85:
Build that executed and converted, publish to a set of manifests: [here]( https://travis-ci.org/heroku/stack-images/builds/493448451)

Any later builds won't be converting or publishing to the cheverny app because we only want tagged releases to funnel into cheverny at this time.

Some of the fields we pass in are not yet used by the cheverny app, but we are going ahead and sending them now. (Signature, Package version list).

This will effectively deprecate the stack-image-tools internal repo. Once approved we will archive that repo and update the readmes in the various places. This will make a prime effort a bit easier with less repos to juggle. We can also then deprecate the credential holder app that the old repo references in favor of the creds being in cheverny alone.

PR #136, opened 2019-03-08
Update postgres packages to version 11

jabrown85:
Update client libraries to use version 11 of postgresql packages.

[Failing build](https://travis-ci.org/heroku/stack-images/builds/503646642) because of the dockerhub issues going on right now. The build itself is passing, just the final image push is failing.


jabrown85:
@ericc572 Correct. Cedar is long gone and not even built from this repo.

PR #140, opened 2019-05-01
Enable ESM updates for cedar-14

edmorley:
Configures the ESM (Extended Security Maintenance) repository for the cedar-14 stack image, so that it continues to receive updates now that Ubuntu 14.04 has reached the end of its standard support lifecycle:
https://blog.ubuntu.com/2019/02/05/ubuntu-14-04-trusty-tahr

The ESM agreement requires that we not distribute images containing ESM updates, so newly generated cedar-14 images will no longer be published to Docker Hub.

The repository credentials will be set via the Travis dashboard, and the APT auth config file into which they are saved is removed from the final stack image.

Fixes [W-6100727](https://gus.my.salesforce.com/a07B0000006KaKwIAK).

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -23,11 +23,23 @@ write_package_list() {
     docker run --rm "$image_tag" dpkg-query --show --showformat='${Package}\n' >> "$output_file"
 }
 
-display "Building $STACK main image"
-docker build --pull --tag "$IMAGE_TAG" "$DOCKERFILE_DIR" | indent
-write_package_list "$IMAGE_TAG" "$DOCKERFILE_DIR"
+if [[ "${STACK}" = "cedar-14" ]]; then
+    display "Building ${STACK} combined image"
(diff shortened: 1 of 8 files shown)


edmorley:
The Travis build will fail for now, until `ESM_USERNAME` and `ESM_PASSWORD` are set via the dashboard (which I'll do once I've had a first pass review).

bin/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
edmorley: This change is so that when testing locally, the script uses the user's preferred bash (such as the up-to-date one installed by homebrew) rather than system bash (which for the case of OS X is bash v3 which doesn't support the `-v` syntax below).

edmorley:
> Is there any concern for the `ESM_EF1B9BA3` key to be here? I know it's a public key but I'm not sure where it comes from. I assume it's useless without ESM creds to gain access to the ESM packages anyway.

The ESM docs (see links in GUS) have a step to pull the key from the public Ubuntu keyserver, which can be accessed by anyone:
https://keyserver.ubuntu.com/pks/lookup?search=esm%40canonical.com&op=index

edmorley:
I've added the credentials to Travis' dashboard and retriggered Travis - is now passing.

edmorley:
And confirmed the Travis deploy condition is successfully blocking the cedar-14 Docker image publish:
https://travis-ci.org/heroku/stack-images/jobs/526951512#L7436

PR #141, opened 2019-07-11
Add version control tools to the heroku-16/18 build images

jmorrell:
I'm mainly opening this to prompt discussion and learn.

After a recent discussion with @freeformz [slack link](https://heroku.slack.com/archives/C1RU92Y1L/p1562871530124200) I found that the go buildpack needs to install mercurial and bazaar version control tools to support `go.mod` apps. 

Relevant PR: https://github.com/heroku/heroku-buildpack-go/pull/345

These seem like tools that should be in the build stack image. Is there a reason they are not there? Has there been previous discussion around this? (search is not bringing things up)

### DIFF
--- a/heroku-16-build/bin/heroku-16-build.sh
+++ b/heroku-16-build/bin/heroku-16-build.sh
@@ -68,6 +68,7 @@ apt-get install -y --force-yes \
     libxslt-dev \
     libyaml-dev \
     libzip-dev \
+    mercurial \
     postgresql-server-dev-11 \
     python-dev \
     ruby-dev \
--- a/heroku-16-build/bin/heroku-16-build.sh
+++ b/heroku-16-build/bin/heroku-16-build.sh
@@ -9,7 +9,9 @@ apt-get install -y --force-yes \
     autoconf \
     bison \
     build-essential \
+    bzr \
     gettext \
+    git \
     libacl1-dev \
     libapparmor-dev \
     libapt-pkg-dev \
--- a/heroku-16-build/installed-packages.txt
+++ b/heroku-16-build/installed-packages.txt
@@ -17,6 +17,7 @@ bsdutils
 build-essential
 bzip2
 bzip2-doc
+bzr
 ca-certificates
 ca-certificates-java
 clang-6.0
@@ -32,6 +33,7 @@ debconf
 debianutils
 dh-python
 diffutils
+dirmngr
 distro-info-data
 dnsutils
 dpkg
@@ -64,6 +66,8 @@ gir1.2-rsvg-2.0
 git
 git-man
 gnupg
+gnupg-agent
+gnupg2
 gpgv
 grep
 gsfonts
@@ -105,6 +109,7 @@ libasan2
 libasn1-8-heimdal
 libasprintf-dev
 libasprintf0v5
+libassuan0
 libatk1.0-0
 libatk1.0-data
 libatm1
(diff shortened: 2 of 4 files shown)


freeformz:
Adding this to the build images would allow me to remove stuff from the go buildpack and generally make things easier for a few customers. So I'm +1000 on this.

freeformz:
Related to this https://github.com/heroku/heroku-buildpack-go/pull/345

jmorrell:
Note: I've not tested this and am mainly interested in learning if there are any blockers / objections. CI is currently failing, but I will do due diligence on the PR if there aren't any.

freeformz:
I fixed the build errors and at least one more.

edmorley:
The Postgres key error fix has been split out to #143 - could you rebase on top of master and remove that change? (The key has been replaced rather than a second added; which is why it's not showing any conflicts)

edmorley:
> These seem like tools that should be in the build stack image. Is there a reason they are not there? Has there been previous discussion around this? (search is not bringing things up)

I like the idea of adding these packages. I don't remember seeing any discussion about them on this repo before.

In general I think the possible reasons for not adding a new package are:
* that it increases the build/runtime image sizes too much - however this PR only results in a 40MB increase for `heroku-16-build` and 20MB for `heroku-18-build` (and none for the runtime image), so a non-issue
* that it allows for anti-patterns (for example including the sqlite headers makes it easier for apps to inadvertently use ephemeral dyno storage for ORM backends) - but I don't think that applies here
* that it might be incompatible with existing buildpacks/apps

Regarding the last point - I'm presuming the Go buildpack will happily continue to use the vendored binaries in preference to the new ones in the stack image, so therefore not an issue?

jmorrell:
@edmorley rebased 👍

tools/bin/capture-package-versions
```diff
@@ -2,7 +2,8 @@
 
 set -e -o pipefail
 
+csd=$(dirname $0)
 DOCKER_IMAGE=$1
 OUTPUT_FILE=$2
 
-docker run --rm -ti $DOCKER_IMAGE dpkg-query -W -f "\${package},\${version}\n" | jq -s -R -f /usr/local/bin/csv-to-array.jq > $OUTPUT_FILE
+docker run --rm -ti $DOCKER_IMAGE dpkg-query -W -f "\${package},\${version}\n" | jq -s -R -f ${csd}/csv-to-array.jq > $OUTPUT_FILE
```
edmorley: This works on Travis due to:
https://github.com/heroku/stack-images/blob/fd80801050221a436108308a7a705d7eaf840a1c/bin/convert-and-publish-to-heroku.sh#L3

I'm presuming this script was being run by hand locally? If so, the way we've been updating the installed-packages list until now is via the `bin/build.sh` script, which doesn't use this file. 

There's definitely some additional clean-up that can be performed now that the stack image tools repo has been merged into this one (of which this is one part) - however since it's unrelated to this PR and not necessary to update the packages list locally, I think we should exclude this change for now.

jmorrell: I'll remove this commit 👍

edmorley:
This has been released to staging, if you'd like to test there? :-)

edmorley:
The changes here are now in production.

edmorley:
Unfortunately today's stack image update had to be rolled back (due other dependency updates from upstream), so the VCS tools won't be available in production for a bit longer.

PR #146, opened 2019-10-01
Add libc-client2007e to the stack image

dzuelke:
for all stacks

this prevents us from having to maintain a custom patched version inside buildpacks (see https://bugs.launchpad.net/ubuntu/+source/uw-imap/+bug/1834340)

### DIFF
--- a/cedar-14/bin/cedar-14.sh
+++ b/cedar-14/bin/cedar-14.sh
@@ -191,6 +191,7 @@ apt-get install -y --force-yes \
     iputils-tracepath \
     language-pack-en \
     libbz2-dev \
+    libc-client2007e \
     libcurl4-openssl-dev \
     libev-dev \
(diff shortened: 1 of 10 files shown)


cedar-14/bin/cedar-14.sh
```diff
@@ -191,6 +191,7 @@ apt-get install -y --force-yes \
     iputils-tracepath \
     language-pack-en \
     libbz2-dev \
+    libc-client2007e \
```
edmorley: Should `libc-client2007e-dev` also be here too, given that Cedar-14's image is combined runtime+buildtime image, and `libc-client2007e-dev` has been added to the build time variants below?

dzuelke: We've generally not done that until really necessary in order to keep the image size reasonable.

PR #149, opened 2019-10-22
add libmcrypt4 to cedar-14

dzuelke:
so the PHP buildpack doesn't have to maintain it by hand for that stack


dzuelke:
Nope that's fine @edmorley, the dev package can be installed during compilation (since it's all done in Docker).

PR #150, opened 2019-11-06
Add cmake package to heroku-16/18 build images

monde:
I'd like to use the [pronto gem](https://github.com/prontolabs/pronto) on one of our internal apps that makes use of the Heroku-16-build image in its testing environment. The pronto gem builds with native bindings including [`libgit2/rugged`](https://github.com/libgit2/rugged) which has a `cmake` dependency. See:
```
# ...
# library versions scrubbed in this example with XXX
       Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

       current directory:
       /app/vendor/bundle/ruby/X.X.X/gems/rugged-X.X.X/ext/rugged
       /app/vendor/ruby-X.X.X/bin/ruby -r ./siteconfXXXvwnd.rb extconf.rb
       checking for gmake... no
       checking for make... yes
       checking for cmake... no
       ERROR: CMake is required to build Rugged.
       *** extconf.rb failed ***
       Could not create Makefile due to some reason, probably lack of necessary
       libraries and/or headers.  Check the mkmf.log file for more details.  You may
       need configuration options.
# ...
```

### DIFF
--- a/heroku-16-build/bin/heroku-16-build.sh
+++ b/heroku-16-build/bin/heroku-16-build.sh
@@ -10,6 +10,7 @@ apt-get install -y --force-yes \
     bison \
     build-essential \
     bzr \
+    cmake \
     gettext \
     git \
     libacl1-dev \
--- a/heroku-16-build/installed-packages.txt
+++ b/heroku-16-build/installed-packages.txt
@@ -21,6 +21,8 @@ bzr
 ca-certificates
 ca-certificates-java
 clang-6.0
+cmake
+cmake-data
 comerr-dev
 coreutils
 cpp
@@ -105,6 +107,7 @@ libapparmor1
 libapt-inst2.0
 libapt-pkg-dev
 libapt-pkg5.0
+libarchive13
 libasan2
 libasn1-8-heimdal
 libasprintf-dev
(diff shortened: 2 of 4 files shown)


jabrown85:
Is there a reason to have `cmake` available on the _runtime_ image? I don't have any objections to including it in `heroku-16-build` but if we don't have a use case for needing it outside of build I would only add it to the build variant.

Additionally, I would propose adding `cmake` to `heroku-18-build` as well. We have historically tried to not have 16/18 drift in packages unless there was a need.

monde:
I will follow @jabrown85's advice re: build image only and in heroku-16 and heroku-18

monde:
I squashed out my original changes to Heroku-16 and added cmake to the build images for Heroku-16 and Heroku-18. Also make @dzuelke the reviewer for this PR.

edmorley:
This PR look good to me with those changes. Adds 20MB to each build image, however we're less concerned about the size of the build image variants, and the cmake package is generic enough that it may be useful for many others. It's also not a package that has security noticed very often (last I can find on usn.ubuntu.com was 2010), so not likely to add the the frequency with which we have to roll the images.

PR #148, opened 2019-10-17
Temporarily pin libpq5 and libpq-dev to v11

edmorley:
Since libpq 12 includes a breaking change (more strict validation of connection parameters) which causes connection errors at runtime for customers with invalid database configuration settings.

This unblocks new stack image releases until an assessment of the number of customers affected/outreach can be performed.

The version is being pinned using the approach here:
https://wiki.postgresql.org/wiki/Apt/FAQ#I_want_libpq5_for_version_X.2C_but_there_is_only_version_Y_in_the_repository

Cedar-14's build script did not need to be updated, since PostgreSQL 12 is not available in the `trusty-pgdg` archive (due to Ubuntu 14.04 EOL).

Refs #147 / [W-6780695](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007azZLIAY/view).

PR #151, opened 2020-01-14
Update postgresql-client, postgresql-server-dev and libpq to PG12

### LINKED ISSUE #147: Breaking changes in PostgreSQL 12's libpq
The stack images currently install the PostrgreSQL client, library and headers from the upstream PostgreSQL APT repository (`apt.postgresql.org`) rather than the less frequently updated packages in Ubuntu's APT repository. 

This means that as new major Postgres versions are released, the next stack image release will include updated major versions of `libpq5` and `libpq-dev` (the clients themselves use versioned package names such as `postgresql-client-11` so have to be explicitly updated).

New major versions of libpq are generally backwards compatible, however libpq v12 includes the following breaking change that makes the validation of connection parameters more strict:
https://github.com/postgres/postgres/commit/e7a2217978d9cbb2149bfcb4ef1e45716cfcbefb

Unfortunately this breaking change wasn't listed in the [PostgreSQL 12 release notes](https://www.postgresql.org/docs/12/release-12.html), so we weren't able to factor it into the stack release review process.

This meant the [latest stack image update](https://devcenter.heroku.com/changelog-items/1688) caused database connection errors at runtime for a couple of customers, since their Rails apps were using invalid values for connection parameters (such as `connect_timeout` having trailing non-numeric characters). (Previously with libpq v11 these invalid values were being silently ignored.)

As such we:
* rolled back that stack image release
* [reported](https://www.postgresql.org/message-id/157123155668.25311.9369950798665566339%40wrigleys.postgresql.org) the missing release notes breaking change entry to the Postgres project
* filed ged/ruby-pg/issues/302 to have the pg gem report the more helpful validation related error message instead of `PG::UnableToSend: no connection to the server`

Determining how to best roll out the PG 12 client libraries will be more involved, so in the short term we should pin to v11 so that stack image releases are unblocked. After that someone will need to investigate further as to how widespread invalid app configurations are and work on a rollout plan.

CC @heroku/dod-infra since they will need to be aware of this if wanting to upgrade to PG12 in the stack image in the future. For more info see the investigation notes [here](https://salesforce.quip.com/8Ik7Abj1T0sK).

edmorley:
This undoes the pinning of `libpq` added in #148 and additionally updates `postgresql-client` and `postgresql-server-dev` to the PG12 releases, to support Heroku Postgres 12.

Since #148 the upstream Postgres project has improved the libpq 12 error handling, such that it now provides a clearer error message when the validation of connection parameters fail, which makes it easier for customers to identify that it's due to their app having an invalid configuration.

The build scripts for Cedar-14 have not been modified since the upstream Postgres project has not released PG12 versions of the packages for Ubuntu-14 given it is EOL.

Release notes:
https://www.postgresql.org/docs/12/release-12.html

Refs [W-7113082](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007sgomIAA/view).
Closes #147.

### DIFF
--- a/heroku-16-build/bin/heroku-16-build.sh
+++ b/heroku-16-build/bin/heroku-16-build.sh
@@ -71,7 +71,7 @@ apt-get install -y --force-yes \
     libyaml-dev \
     libzip-dev \
     mercurial \
-    postgresql-server-dev-11 \
+    postgresql-server-dev-12 \
     python-dev \
     ruby-dev \
     zlib1g-dev \
--- a/heroku-16-build/installed-packages.txt
+++ b/heroku-16-build/installed-packages.txt
@@ -541,10 +541,10 @@ pgdg-keyring
 pinentry-curses
 pkg-config
 poppler-data
-postgresql-client-11
+postgresql-client-12
 postgresql-client-common
 postgresql-common
-postgresql-server-dev-11
+postgresql-server-dev-12
 procps
 python
 python-bzrlib
--- a/heroku-16/bin/heroku-16.sh
+++ b/heroku-16/bin/heroku-16.sh
@@ -9,17 +9,7 @@ deb http://archive.ubuntu.com/ubuntu/ xenial main universe
 deb http://archive.ubuntu.com/ubuntu/ xenial-security main universe
 deb http://archive.ubuntu.com/ubuntu/ xenial-updates main universe
 
-# Use the PG11 version of libpq to work around:
-# https://github.com/heroku/stack-images/issues/147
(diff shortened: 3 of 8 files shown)


jabrown85:
Should heroku/heroku-buildpack-ci-postgresql/pull/29 be merged when this goes live? Or is this PR going to go in before PG12 is the default for Heroku?

edmorley:
@jabrown85 This stack image change can occur at any point prior to Heroku Postgres 12 GA - it doesn't need to be coordinated with it.

PR #152, opened 2020-01-30
add libonig to heroku-16 and heroku-18

dzuelke:
regular expressions library, now required by PHP 7.4 for multibyte strings


dzuelke:
(install size of runtime library is ~500 kB)

PR #154, opened 2020-04-09
Add Heroku-20 stack

edmorley:
This is an initial pass at creating a new Heroku-20 stack, based on Ubuntu 20.04 ("Focal Fosa").

Ubuntu 20.04 isn't yet GA ([release schedule](https://wiki.ubuntu.com/FocalFossa/ReleaseSchedule)), but this PR is just the first of many steps required to support a new stack (see [here](https://salesforce.quip.com/LbhlAFPd1rvR) for a rough sketch of the rest), so it would be good to start the discussion now, even if we don't have a confirmed roadmap.

The process for creating the new stack was:
1. Copy the `heroku-18{,-build}` directories to `heroku-20{,-build}`.
2. Adjust base image / APT sources to substitute 18.04/bionic references with 20.04/focal.
3. Adjust the packages list to account for renamed/removed packages, until the build passed.
4. Update documentation/Travis files.

Fixes [W-7429923](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007QKjyIAG/view).

**When reviewing look at the individual commits**, since I've split the file copy and adjustments up to make it possible to see what's changed compared to Heroku-18.

Of note:
* The `python` and `python-dev` packages (which were the Python 2 versions) no longer exist under Ubuntu 20.04. However there are still explicitly named `python2` and `python2-dev` packages - so it is at least possible to install Python 2 (if we want).
* Even if we don't explicitly install the `python2` package (this initial PR doesn't, it only explicitly installs Python 3), for the build image (only) it's pulled in by other packages regardless.
* The `python2` and `python2-dev` packages no longer install a binary/symlink for `python`. ie: Users must explicitly call the command `python{2,3}` and not `python`, unless we choose to add it back for compat (using the `﻿﻿python-is-python2` or `python-is-python3` packages).

For testing, the Travis run for this PR will upload test images to these Docker Hub tags:
* `heroku/heroku-20.nightly`
* `heroku/heroku-20-build.nightly`

Open questions:
1. Should we explicitly install `python2` in the runtime image, for parity with the build image? (It gets pulled in for the build image whether we like it or not)
2. Should we use `﻿﻿python-is-python2` or `python-is-python3` (that symlink `python`) for compat?
3. Whilst we're releasing the new stack, are there any other breaking changes we might want to make? (With the caveat that every breaking change is another barrier when we come to asking customers to migrate from earlier stacks for their EOL - so I'm leaning towards Heroku-20 being as similar to Heroku-18 as possible.)

For (1) and (2) we can always add them later (whereas we can't remove once GA), so we could perhaps start without and see how we go?


dmathieu:
FTR: 

> heroku/heroku:20         553MB
> heroku/heroku:20-build   1.41GB

edmorley:
Thank you for the review :-)

Any other takers to review, before I merge tomorrow? (We will no doubt iterate on the new stack prior to release, so no massive rush)

edmorley:
> Open questions:
> ...
> 2. Should we use ﻿﻿python-is-python2 or python-is-python3 (that symlink python) for compat?

`python-is-python3` is being added in #182.

PR #162, opened 2020-04-29
Run shellcheck in CI

edmorley:
Shellcheck lints shell scripts, helping to prevent common mistakes:
https://github.com/koalaman/shellcheck#gallery-of-bad-code

It would have caught the typo that was fixed in #161.

In order for shellcheck to pass, the following needed fixing:

- ```SC2006: Use $(...) notation instead of legacy backticks `...`.```
          -> https://www.shellcheck.net/wiki/SC2006
- `SC2046: Quote this to prevent word splitting.`
          -> https://www.shellcheck.net/wiki/SC2046
- `SC2048: Use "$@" (with quotes) to prevent whitespace problems.`
          -> https://www.shellcheck.net/wiki/SC2048
- `SC2068: Double quote array expansions to avoid re-splitting elements.`
          -> https://www.shellcheck.net/wiki/SC2068
- `SC2086: Double quote to prevent globbing and word splitting.`
          -> https://www.shellcheck.net/wiki/SC2086

All non-sourced scripts have also been updated to use stricter bash modes:
http://redsymbol.net/articles/unofficial-bash-strict-mode/

These changes were tested by manually triggering another Heroku-20 Docker+stack image release on an adhoc branch (since PRs/master don't run all steps):
https://travis-ci.com/github/heroku/stack-images/jobs/324852938

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - STACK=heroku-18 IMAGE_TAG=heroku/heroku:18 PRIVATE_IMAGE_TAG=heroku/heroku-private:18
   - STACK=heroku-20 IMAGE_TAG=heroku/heroku:20 PRIVATE_IMAGE_TAG=heroku/heroku-private:20
 script:
+- find . -type f \( -name "*.sh" -o -path "*/bin/*" \) ! -name '*.jq' | xargs -t shellcheck
 - bin/build.sh $STACK $IMAGE_TAG $IMAGE_TAG-build
 - |
   status="$(git status --porcelain)"
@@ -22,7 +23,7 @@ deploy:
     on:
       tags: true
   - provider: script
-    script: sh bin/publish-to-dockerhub.sh
+    script: bin/publish-to-dockerhub.sh
     on:
       # Cedar-14 images must not be published to Docker Hub since they contain ESM updates.
       condition: ($TRAVIS_BRANCH == "master" || -n $TRAVIS_TAG) && $STACK != "cedar-14"
--- a/bin/stack-helpers.sh
+++ b/bin/stack-helpers.sh
@@ -1,11 +1,12 @@
-set -o pipefail
+# shellcheck shell=bash
 
 function display() {
-  echo -e "\n----->" $*
+  echo -e "\n-----> $1"
 }
 
 function abort() {
-  echo $* ; exit 1
+  echo "$1"
+  exit 1
 }
 
 function indent() {
--- a/tools/bin/make-filesystem-image
+++ b/tools/bin/make-filesystem-image
@@ -1,10 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-set -e
+set -euo pipefail
 
 IMG="$1"
 
 mkdir -p "$(dirname "$IMG")"
 dd if=/dev/zero of="$IMG" bs=100M count=24
-yes | mkfs -t ext3 -m 1 "$IMG"
+mkfs -t ext3 -m 1 "$IMG"
 tune2fs -c 0 -i 0 "$IMG"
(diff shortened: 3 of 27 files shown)


tools/bin/make-filesystem-image
```diff
@@ -1,10 +1,10 @@
...
+#!/usr/bin/env bash
 
-set -e
+set -euo pipefail
 
 IMG="$1"
 
 mkdir -p "$(dirname "$IMG")"
 dd if=/dev/zero of="$IMG" bs=100M count=24
-yes | mkfs -t ext3 -m 1 "$IMG"
```
edmorley, line 9: The `yes` command outputs `y` strings repeatedly until it is killed by the `SIGPIPE` when mkfs exits. Its non-clean exit causes the overall script to error due to `pipefail` mode. 

We could disable `pipefail` for this command only, however the use of `yes` was a [workaround](https://askubuntu.com/questions/428474/automatic-say-yes-to-make-ext3-file-system-on-disk-with-mkfs-ext3) for `mkfs` needing a confirmation prompt in certain cases - and trying mkfs now without the `yes` shows it no longer prompts, so it's redundant.

(I tracked the first usage of `yes` with `mkfs` to: https://github.com/heroku/capturestack/commit/ac525cac9291aa376f16e6f5ac460c9a77cf4d2f - that repo was presumably the inspiration for https://github.com/heroku/stack-image-tools/pull/3 which was then eventually vendored here.)

bin/stack-helpers.sh
```diff
@@ -1,11 +1,12 @@
-set -o pipefail
+# shellcheck shell=bash
```
edmorley, line 1: This is a helper script that is sourced, so doesn't need a shebang/bash mode setting.

.travis.yml
```diff
@@ -22,7 +23,7 @@ deploy:
     on:
       tags: true
   - provider: script
-    script: sh bin/publish-to-dockerhub.sh
```
edmorley, line 25: Means the script is run with what's specified in the shebang (bash) and not `sh`.

.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
   - STACK=heroku-18 IMAGE_TAG=heroku/heroku:18 PRIVATE_IMAGE_TAG=heroku/heroku-private:18
   - STACK=heroku-20 IMAGE_TAG=heroku/heroku:20 PRIVATE_IMAGE_TAG=heroku/heroku-private:20
 script:
+- find . -type f \( -name "*.sh" -o -path "*/bin/*" \) ! -name '*.jq' | xargs -t shellcheck
```
edmorley, line 12: Can't use only bash globbing (eg `shellcheck **/bin/* **/*.sh`) since otherwise shellcheck will be passed the same filename twice, and currently the input file list is not de-duplicated, so any failures would show up in the results twice:
https://github.com/koalaman/shellcheck/issues/1930

bin/stack-helpers.sh
```diff
@@ -1,11 +1,12 @@
...
 
 function display() {
-  echo -e "\n----->" $*
+  echo -e "\n-----> $1"
```
jabrown85, line 4: What is the story with this change?

edmorley: Ah so shellcheck warns on usage of unquoted `$*` since it applies both word splitting and globbing:
https://github.com/koalaman/shellcheck/wiki/SC2048

The error could have been resolved by changing that line to:
```sh
   echo -e "\n-----> $*"
```

...however it seemed cleaner to have the API of `display` and `abort` be that they are passed quoted strings, to avoid issues with word splitting, and also so that it's clearer when viewing in an editor with syntax highlighting, that they are message strings and not arguments to a command.

Happy to change it back if preferred :-)

jabrown85: That is totally fine. I just noticed it _looked_ like it would change behavior. As long as we are confident we didn't use it in that way I'm 👍  cleaning up the API here to provide some safety and symmetry.

PR #163, opened 2020-04-30
Stop using --force-yes with apt-get on newer stacks

edmorley:
Since as of APT 1.1 (Ubuntu 16.04+) it results in a deprecation warning:

```
W: --force-yes is deprecated, use one of the options starting with --allow instead.
```
(Example: https://travis-ci.com/github/heroku/stack-images/jobs/325509026#L1794)

It also appears to no longer be required, since the build completes fine in non-interactive mode without it, and no additional prompts are seen even if the build is run by hand interactively.

I've also confirmed no differences in output (other than the expected changes in log files/caches), using:
`container-diff diff --type=file <image-before> <image-after>`

See also:
https://manpages.ubuntu.com/manpages/xenial/en/man8/apt-get.8.html

### DIFF
--- a/heroku-16-build/setup.sh
+++ b/heroku-16-build/setup.sh
@@ -9,7 +9,7 @@ export DEBIAN_FRONTEND=noninteractive
 export LC_ALL=C
 
 apt-get update
-apt-get install -y --force-yes \
+apt-get install -y \
     autoconf \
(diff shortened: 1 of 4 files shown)


edmorley:
Thank you for the review(s) :-)

PR #166, opened 2020-05-12
Add the source git reference to the image metadata API call

edmorley:
So that it can be tracked/used by the app that manages image manifests, now that it supports (and requires) it be sent in the API call:
https://github.com/heroku/cheverny/pull/62

The `TRAVIS_TAG` and `TRAVIS_BRANCH` env vars are documented here:
https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

Refs [W-7531996](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007obh6IAA/view).


edmorley:
This has been tested on staging:
https://github.com/heroku/cheverny/pull/62#issuecomment-627456101

PR #179, opened 2020-09-08
Add libc6-dev into the stack image

schneems:
Currently on heroku-18 and heroku-20 you cannot compile at runtime, but can on heroku-16. This prevents `mjit` which relies on out of band compilation to implement jit. Here are some details about how I uncovered this:

On a heroku run bash instance this code fails on heroku-18 but works on heroku-16:

```
~ $ cat > numprocessors.c <<EOL
>   #include <unistd.h>
>   #include <stdio.h>
>
>   int main() {
>     int val = sysconf(_SC_NPROCESSORS_ONLN);
>     printf("# => Number of processors from sysconf: %i\n", val);
>     return 0;
>   }
> EOL
~ $ gcc numprocessors.c && ./a.out
numprocessors.c:1:12: fatal error: unistd.h: No such file or directory
   #include <unistd.h>
            ^~~~~~~~~~
```

here's a blog post from when it worked https://schneems.com/2017/08/01/is-webrick-webscale/

Also Ruby's "mjit" does not work on Heroku-18:

```
~ $ ruby --jit-wait --jit-verbose=5 -e "puts 'yoyoyoyo'"
MJIT: CC defaults to /usr/bin/gcc
MJIT: tmp_dir is /tmp
Locking in mjit_cont_new
Locked in mjit_cont_new
Click to expand inline (14 lines)
```

It deadlocks since it's never able to compile anything using mjit, it prints those unlock/lock lines out forever.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,5 +1,24 @@
 # Building Stack Images Locally
 
+## Prepare your local environment
+
+The build scripts in this repository require bash 4+. To update to newer bash on OS X, see:
+https://johndjameson.com/blog/updating-your-shell-with-homebrew/
+
+## Adding packages to the stack image
+
+Add the package you want to the approproate `setup.sh` for example `heroku-18/setup.sh`:
+
+```diff
++    libc6-dev \
+```
+
+Once done, run the `bin/build.sh` locally to generate the corresponding `installed-packages.txt`.
+
+The `*-build` variants include all the packages from the non-build variant by default. This means that if you're adding a package to both, you only need to add them to the non-build variant. The example above will add `libc6-dev` to both `heroku-18` and `heroku-18-build`.
+
+## Build
+
 To build the stack images locally, run this from the repo root:
 
     bin/build.sh STACK DOCKER_TAG DOCKER_BUILD_TAG
--- a/heroku-18/installed-packages.txt
+++ b/heroku-18/installed-packages.txt
@@ -87,7 +87,9 @@ libbsd0
 libbz2-1.0
 libc-bin
 libc-client2007e
+libc-dev-bin
 libc6
+libc6-dev
 libcairo2
 libcap-ng0
 libcap2
@@ -279,6 +281,7 @@ libxslt1.1
 libyaml-0-2
 libzip4
 libzstd1
+linux-libc-dev
 locales
 login
 lsb-base
--- a/heroku-18/setup.sh
+++ b/heroku-18/setup.sh
@@ -133,6 +133,7 @@ apt-get install -y --no-install-recommends \
     less \
     libargon2-0 \
     libc-client2007e \
+    libc6-dev \
     libcairo2 \
     libcroco3 \
     libcurl4 \
--- a/heroku-20/installed-packages.txt
+++ b/heroku-20/installed-packages.txt
@@ -89,7 +89,9 @@ libbsd0
 libbz2-1.0
 libc-bin
 libc-client2007e
+libc-dev-bin
 libc6
+libc6-dev
 libcairo-gobject2
 libcairo2
 libcap-ng0
@@ -99,6 +101,7 @@ libcbor0.6
 libcc1-0
 libcom-err2
 libcroco3
+libcrypt-dev
 libcrypt1
 libctf-nobfd0
 libctf0
@@ -282,6 +285,7 @@ libxtables12
 libyaml-0-2
 libzip5
 libzstd1
+linux-libc-dev
 locales
 login
 logsave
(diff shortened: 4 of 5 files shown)


BUILD.md
```diff
@@ -1,5 +1,21 @@
 # Building Stack Images Locally
 
+## Prepare your local environment
+
+the `bin/build.sh` is written in bash and requires bash 4+. On mac here is how you can update your bash version https://johndjameson.com/blog/updating-your-shell-with-homebrew/.
```
edmorley, line 5:
```suggestion
The build scripts in this repository require bash 4+. To update to newer bash on OS X, see:
https://johndjameson.com/blog/updating-your-shell-with-homebrew/
```

heroku-18-build/setup.sh
```diff
@@ -26,6 +26,7 @@ apt-get install -y --no-install-recommends \
     libaudit-dev \
     libbsd-dev \
     libbz2-dev \
+    libc6-dev \
```
edmorley, line 29: The `*-build` image variants use the non-build images as their base. As such only build-specific packages need to be listed in `*-build/setup.sh`, since the others will already be installed.
```suggestion
```

heroku-20-build/setup.sh
```diff
@@ -27,6 +27,7 @@ apt-get install -y --no-install-recommends \
     libbsd-dev \
     libbz2-dev \
     libc-client2007e-dev \
+    libc6-dev \
```
edmorley, line 30:
```suggestion
```

edmorley (changes requested):
Thank you for adding this! :-)

> here's a blog post from when it worked https://schneems.com/2017/08/01/is-webrick-webscale/

The `installed-packages.txt` files in this repo ensure that we never accidentally remove a package from a stack after it's released. As such, it's most likely that mjit has never worked on Heroku-18+ rather than it having stopped working at some point.

The blog post was from 2017, prior to the release of Heroku-18 and when Heroku-16 was the default stack (https://devcenter.heroku.com/changelog-items/1139), so I'm presuming this is why it worked out of the box before.

BUILD.md
```diff
@@ -1,5 +1,22 @@
...
+
+The build scripts in this repository require bash 4+. To update to newer bash on OS X, see:
+https://johndjameson.com/blog/updating-your-shell-with-homebrew/
+
+## Adding packages to the stack image
+
+Add the package you want to the approproate `setup.sh` for example `heroku-18-build/setup.sh`:
+
+```diff
++    libc6-dev\
```
edmorley, line 13: (Sorry missed this before)
```suggestion
+    libc6-dev \
```

edmorley (approved):

schneems:
> The blog post was from 2017, prior to the release of Heroku-18 and when Heroku-16 was the default stack (https://devcenter.heroku.com/changelog-items/1139), so I'm presuming this is why it worked out of the box before.

Right, and I don't think anyone noticed. I can add an mjit test in the Ruby buildpack, but it seems pretty far away from this repo. Is there a way I could test this here? Maybe my small `gcc` example?

> The *-build image variants use the non-build images as their base. As such only build-specific packages need to be listed in *-build/setup.sh, since the others will already be installed.

I'm going to add docs for this as well then. I'm re-running `bin/build` after our updates. I'm curious about the `installed-packages.txt` here. Is it assumed that the *-build/installed-packages.txt` also contains the packages of the non-build so they don't need to be updated here?

edmorley:
> Is there a way I could test this here? Maybe my small gcc example?

I think it would definitely be useful if the CI checked more than just "does this image build" + "does the installed packages list match the fixture" - since it would reduce iteration time for PRs here and also mean the CI run on tag (ie stack image release) catches more issues (rather than post release on Direwolf staging etc).

That said, I think it would be best if we didn't duplicate too much here - perhaps we could (eventually) have the CI here run Hatchet tests or getting started guides from all of the official languages?

> I'm curious about the installed-packages.txt here. Is it assumed that the *-build/installed-packages.txt` also contains the packages of the non-build so they don't need to be updated here?

The `installed-packages.txt` file is a dump of every installed package in that image (see https://github.com/heroku/stack-images/blob/b2c9ca3a1b6ca979401f4c1783167f04475cc484/bin/build.sh#L22-L23). The build images package list will be the packages that existed in the base image (the non-build image) + whatever was installed on top in the build image.

edmorley:
> That said, I think it would be best if we didn't duplicate too much here

To be clearer: I think the canonical location for these types of language-specific tests should still be in the buildpack's repo, but perhaps we could have CI here clone that and run tests or ...?

PR #180, opened 2020-09-08
Add `vi` to the stack image

schneems:
When debugging customer issues in production it can be necessary to modify files with debugging statements. Currently there are no editors on runtime images. This PR adds the basic `vi` to the runtime image. 

```
docker run -it --rm heroku/heroku:18 bash
root@c835b7f1d03d:/# vi --version | head -n 1
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar 18 2020 18:29:15)
```

This appears to add 2mb to the stack image:

```
       heroku/heroku:18                                     487MB # main
       heroku/heroku:18                                     489MB # tiny-vim
```

## Add a package checklist

- How likely is this to break existing builds?
  - Very unlikely
- How much of an impact will it have on stack image size?
  - 2mb
-  Is is covered by standard security update policy?
   - It is in the Main respository: https://ubuntuupdates.org/package_metas?q=vim-tiny
- How often does it have a security issue?
  - Not that often https://ubuntuupdates.org/package/core/xenial/main/security/vim-tiny
-  Is it for build only (in which case a bit more flexible) or runtime too?
   - vi is for runtime only 
- Does it encourage any anti-patterns (eg libsqlite3-dev)
  - Being able to edit files in prod might give people the mistaken impression that this is like the bad old days of ssh where you could pop into a server and do a "hot fix" but modifications will be restricted to only the bash dyno. This is not that bad.

### DIFF
--- a/heroku-16-build/installed-packages.txt
+++ b/heroku-16-build/installed-packages.txt
@@ -618,6 +618,8 @@ ubuntu-keyring
 ucf
 unzip
 util-linux
+vim-common
+vim-tiny
 wget
 x11-common
 x11proto-core-dev
--- a/heroku-16/installed-packages.txt
+++ b/heroku-16/installed-packages.txt
@@ -362,6 +362,8 @@ ubuntu-keyring
 ucf
 unzip
 util-linux
+vim-common
+vim-tiny
 wget
 xauth
 xdg-user-dirs
--- a/heroku-16/setup.sh
+++ b/heroku-16/setup.sh
@@ -155,6 +155,7 @@ apt-get install -y \
     tar \
     telnet \
     tzdata \
+    vim-tiny \
     wget \
     zip \
 
--- a/heroku-18-build/installed-packages.txt
+++ b/heroku-18-build/installed-packages.txt
@@ -563,6 +563,8 @@ ubuntu-keyring
 ucf
 unzip
 util-linux
+vim-common
+vim-tiny
 wget
 x11-common
 x11proto-core-dev
@@ -570,6 +572,7 @@ x11proto-dev
 x11proto-xext-dev
 xorg-sgml-doctools
 xtrans-dev
+xxd
 xz-utils
 zip
 zlib1g
--- a/heroku-18/installed-packages.txt
+++ b/heroku-18/installed-packages.txt
@@ -347,7 +347,10 @@ ubuntu-keyring
 ucf
 unzip
 util-linux
+vim-common
+vim-tiny
 wget
+xxd
 xz-utils
 zip
 zlib1g
--- a/heroku-18/setup.sh
+++ b/heroku-18/setup.sh
@@ -196,6 +196,7 @@ apt-get install -y --no-install-recommends \
     telnet \
     tzdata \
     unzip \
+    vim-tiny \
     wget \
     xz-utils \
     zip \
--- a/heroku-20-build/installed-packages.txt
+++ b/heroku-20-build/installed-packages.txt
@@ -576,6 +576,8 @@ ucf
 unzip
 util-linux
 uuid-dev
+vim-common
+vim-tiny
 wget
 x11-common
 x11proto-core-dev
@@ -583,6 +585,7 @@ x11proto-dev
 x11proto-xext-dev
 xorg-sgml-doctools
 xtrans-dev
+xxd
 xz-utils
 zip
 zlib1g
--- a/heroku-20/installed-packages.txt
+++ b/heroku-20/installed-packages.txt
@@ -347,7 +347,10 @@ ubuntu-keyring
 ucf
 unzip
 util-linux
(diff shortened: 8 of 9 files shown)


jabrown85:
I think I am personally 👎  this b/c of `heroku exec` or `ps:exec` 

> I edited the file and now it isn't working today!

BUT, I think I would be OK with it on _just_ the build stack. It could be very useful for `heroku ci:debug`

I'm curious to hear what others think though!

schneems:
You can already footgun yourself way worse in `ps:exec` without this, `rm -rf /app` and you're toast. I think people either accept that `ps:exec` is a risky place to be, or if not then we need to ask ourselves how we can make it more obvious. For me if I cannot get this on a normal runtime then it's not really useful to me. I've used vim on a `ci:debug` issue before, but probably only once or twice for every 100 "regular" tickets that I've worked on. My main motivation here is decreasing support ticket debug time. 

There are some cases where you want to ask a customer "can you try this different config and see if you get the same error" and that question/answer cycle takes a full 24 hours when otherwise I would just be able to check myself (assuming access permission given). While yes, I can pull vim in remotely it's a pain to do and as a result means I use it less than I would if it was present at debug time.

dmathieu:
I'm also 👎  to this. Making it easier for customers to edit their files manually at runtime may cause another kind of support tickets which would be just as tedious to investigate.

How about an alternative: a CLI plugin introducing a new command such as `heroku run:debug` command which would start a one-of and install VIM on it (we can store it on S3 and use curl to download it) before letting you meddle with it?

Then, we'd have the same behavior: one-off dynos with VIM. But it wouldn't be there on all dynos.

schneems:
> a CLI plugin introducing a new command such as heroku run:debug

Then someone needs to maintain that third party VIM binary for curling. I'm already secretly hosting a statically compiled vim on S3 that I can curl from, but it's on my own time. The stack image maintenance is staffed. A custom CLI plugin is not.

joshwlewis:
An alternative: https://github.com/naaman/heroku-vim. I haven't used it in a few years, but it was fine at the time.

dmathieu:
It's still a ruby CLI plugin. That will definitely not work with the current CLI.
I still think this is the best approach though.

Malax:
I usually download a statically linked `nano` in such cases: https://github.com/andrew-d/static-binaries

schneems:
This sounds like an awful lot of individuals working around a very vague definition of "safer" to make it harder to get work done.

We think in terms of what Heroku provides for you, but there are also lots of cases where the platform makes things (that have very valid use cases) hard or impossible. I believe this is such a use case. There's no reason you need a PHD in Heroku to be able to edit a file on a server that you own.

jabrown85:
I wish we had numbers to help drive a decision here.

How often does the lack of vim come up from customers? vs employees?
How often do those users feel so strongly about it they install it via a buildpack or some other solution?
How many tickets do we get where a file _has_ been modified in a running dyno only for the customer to be broken and confused after the next dyno cycle?

The primary concern I have is that once something goes on the stack, it never really comes off. If installing vim ends up eventually creating support burden, we won't be able to simply remove vim 8 months from now.

For me, I've never once needed or wanted to vim inside a customer's dyno for a ticket. That is obviously not the case for language owners. But that is purely anecdotal.

I am not willing to die on this hill, I just want to air the concerns I have. I still lean toward not installing it, but that is just my opinion and is not rooted in anything concrete.

schneems:
> I wish we had numbers to help drive a decision here.

I searched in our support interface and I see 158 tickets that pattern match to "vim" though this also includes someone's name, so the "real" times are fewer than that.

> once something goes on the stack, it never really comes off

I think we can take it off anytime we want. It's not like people are going to compile a native extension and link to it and if they do...I don't think that's a terribly good idea for them.

I understand that this is a normal concern and generally agree with you. I'm also comfortable as one of the current stack owners as part of dyno-packaging to make some wrong decisions and take the responsibility if they blow up.

jkutner:
tl;dr: I don't support adding `vim` to the run-image stack.

There are plenty of customers who would be alarmed by having this available to their developers at run time. It's essentially a question of Trust. Many of our customers want their run-time stack to have as few packages as possible. I've even seen Java customers that want us to remove `ssh` because they don't use it.

As we move toward Cloud Native Buildpacks, we can make it easier for customers to opt-in to putting `vim` and other tools on their stack via [mixins](https://github.com/buildpacks/rfcs/pull/112). But until that time we should be cautious when adding things to the stack. I agree that it's hard to remove them once they are there (at a minimum we need to give users a long runway). 

Ultimately, putting `vim` on the stack would violate the 12-factor app principles. And I think we should approach anything like that with raised a eyebrow.

edmorley:
Closing for now in favour of the internal RFC discussion.

PR #182, opened 2020-10-01
python-is-python3

dzuelke:
This restores 'python' as an available binary, aliased to 'python3'.

We deem this the better alternative to not having it, because programs might invoke 'python', and a 'command not found' error would presumably much more confusing than a Python 3 related error (and those should be getting pretty rare now anyway). Same for anything that has a '/usr/bin/env python' or similar shebang.

Previous stacks have 'python' aliased to 'python2', but given how that's (finally) EOL, we'd only be kicking painful migrations down the road, and people upgrading stacks are already touching their code base anyway, so better to rip the bandaid off in one go.

GUS-W-8140606

### DIFF
--- a/heroku-20/setup.sh
+++ b/heroku-20/setup.sh
@@ -184,8 +184,8 @@ apt-get install -y --no-install-recommends \
     openssh-server \
     patch \
     postgresql-client-12 \
-    python3 \
     python-is-python3 \
+    python3 \
     rename \
     rsync \
     ruby \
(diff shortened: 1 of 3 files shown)


dkvdm:
I'm Dennis Kromhout van der Meer and I approve of this message.

heroku-20/setup.sh
```diff
@@ -185,6 +185,7 @@ apt-get install -y --no-install-recommends \
     patch \
     postgresql-client-12 \
     python3 \
+    python-is-python3 \
```
edmorley, line 188:
```suggestion
    python-is-python3 \
    python3 \
```

(For consistent package sort order with the generated files)

edmorley (approved):
Agree; thank you :-)

For anyone else following this, there's more back-story about the Ubuntu 20.04 changes Python-related changes in #154.

schneems (approved):

PR #184, opened 2020-10-12
add libwebp-dev runtime deps

dzuelke:
We already have libwebp6 on heroku-18 and heroku-20, and libwebpdemux3 on heroku-20, as these are dependencies of other packages.

When installing libwebp-dev during a build, this will pull down libwebpdemux3 (and also libwebpmux3 on heroku-18), as these are dependencies. A program being built may now link against these additional libraries, but they're not in the runtime image.

Explicitly adding all three libraries so the setup list makes more sense to the casual observer.

GUS-W-8202718

### DIFF
--- a/heroku-20-build/installed-packages.txt
+++ b/heroku-20-build/installed-packages.txt
@@ -95,7 +95,6 @@ language-pack-en
 language-pack-en-base
 less
 lib32gcc-s1
-lib32gcc1
 lib32stdc++6
 libacl1
 libacl1-dev
@@ -443,6 +442,7 @@ libuv1-dev
 libvpx-dev
 libvpx6
 libwebp6
+libwebpdemux2
 libwebpmux3
 libwind0-heimdal
 libwmf-dev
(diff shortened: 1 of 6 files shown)


heroku-20-build/installed-packages.txt
```diff
@@ -95,7 +95,6 @@ language-pack-en
 language-pack-en-base
 less
 lib32gcc-s1
-lib32gcc1
```
dzuelke, line 98:
```
# apt-get upgrade
Reading package lists... Done
Building dependency tree       
Calculating upgrade... Done
The following package was automatically installed and is no longer required:
  lib32gcc1
Use 'apt autoremove' to remove it.
The following packages will be upgraded:
…

# dpkg -L lib32gcc1
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/lib32gcc1
```

edmorley (approved):
Good spot, thank you :-)

PR #207, opened 2022-02-10
Add liblzf and zstd packages

dzuelke:
GUS-W-10679040


edmorley (approved):
I'm excited for us to start using zstd more in the future!

Some background for anyone curious:
https://github.com/facebook/zstd
https://engineering.fb.com/2016/08/31/core-data/smaller-and-faster-data-compression-with-zstandard/
https://engineering.fb.com/2018/12/19/core-data/zstandard/

Issue #203, opened 2021-08-26
ImageMagick "area" resource limit should be in pixels ?


MathieuDerelle:
it is currently defined in bytes
https://github.com/heroku/stack-images/blob/e65a994d5540840ebf173528875db4bad5b29467/heroku-18/setup.sh#L213

the documentation is not really clear https://imagemagick.org/script/resources.php

but it looks like that memory/disk resources expect "bytes" and height/width/area expect "pixels"

> Define arguments for the MAGICK_MEMORY_LIMIT, MAGICK_DISK_LIMIT, and MAGICK_MEMORY_LIMIT environment variables with SI prefixes (.e.g 100MB). MAGICK_WIDTH_LIMIT, MAGICK_HEIGHT_LIMIT and MAGICK_AREA_LIMIT accepts pixel suffixes such as MP for mega-pixels (e.g. 100MP)

which would make more sense

dzuelke:
Fixed in #209 for heroku-22

edmorley:
See:
https://github.com/heroku/stack-images/pull/185#issuecomment-1143478370

PR #209, opened 2022-03-05
Heroku-22

dzuelke:
Almost no changes over heroku-20 right now; only bumped `libwebp` and sorted out a few package versions or name changes (e.g. `mime-support` to `media-types`).

Removed `postgresql-server-dev-14`; `libpq-dev` is enough since nobody will be building Postgres extensions.

`libheif1` is now an explicitly installed package, with headers.

ImageMagick policy hasn't changed since 20.04.

To discuss:
- Ruby removal?
- python2 removal?
- add libssl1.1 runtime library?

GUS-W-10340283


edmorley:
> Ruby removal?

Yeah I'd like to remove system Ruby from the stack image, since:
- it's been the cause of confusion and support tickets for users who've intended to use Ruby from the Ruby buildpack, but have had invalid buildpack configs or binstubs (and so have actually inadvertently been using system Ruby, which is often not clear from the resultant error messages)
- aside from Python (which is used by Ubuntu, plus needed for build scripts for many languages; eg node-gyp etc) we don't include any other languages in the base image,
- dropping it reduces image size plus security surface area in terms of things that may cause us to need to do an urgent stack image release.

Worst case we can add it back later if needed.

> python2 removal?

Yeah I'm happy to remove Python 2 (or more, not add it back given it's not in there as-is). People who need Python 2 for system Python can use `heroku-20` in the meantime.

> add libssl1.1 runtime library?

It seems there isn't a `libssl1.1-dev` package available (like there was `libssl1.0-dev` for 1.0), so it means the most we can do is provide the `libssl1.1` runtime library. As such, unless people are attempting to run assets compiled on an older stack on Heroku-22, the runtime library on it's own won't be any use? As such, perhaps we don't include it for now, unless we hear of a use-case that really does need it?

dzuelke:
Yup agreed on all points @edmorley. I forgot to mention it'll likely be useless to include libssl1.1 for the reason you mentioned; also note that in bionic, you could not install headers for both versions, since the header file paths are identical and the packages thus conflict.

heroku-22/setup.sh
```diff
@@ -0,0 +1,240 @@
...
+  <policy domain="resource" name="width" value="16KP"/>
+  <policy domain="resource" name="height" value="16KP"/>
+  <policy domain="resource" name="area" value="128MB"/>
+  <policy domain="resource" name="disk" value="1GiB"/>
+  <policy domain="delegate" rights="none" pattern="URL" />
+  <policy domain="delegate" rights="none" pattern="HTTPS" />
+  <policy domain="delegate" rights="none" pattern="HTTP" />
+  <policy domain="path" rights="none" pattern="@*"/>
+  <policy domain="cache" name="shared-secret" value="passphrase" stealth="true"/>
+</policymap>
```
edmorley, line 223: I was curious to see how this compared to the latest default `/etc/ImageMagick-6/policy.xml` file.

Checking `ubuntu:22.04` the default contents are:

<details>
  <summary>Click to expand</summary>

```
$ docker run --rm -it ubuntu:22.04 bash
...
root@02edb81c402c:/# apt-get update && apt-get install imagemagick-6-common
...
root@02edb81c402c:/# cat /etc/ImageMagick-6/policy.xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE policymap [
  <!ELEMENT policymap (policy)*>
  <!ATTLIST policymap xmlns CDATA #FIXED ''>
  <!ELEMENT policy EMPTY>
  <!ATTLIST policy xmlns CDATA #FIXED '' domain NMTOKEN #REQUIRED
    name NMTOKEN #IMPLIED pattern CDATA #IMPLIED rights NMTOKEN #IMPLIED
    stealth NMTOKEN #IMPLIED value CDATA #IMPLIED>
]>
<!--
  Configure ImageMagick policies.

  Domains include system, delegate, coder, filter, path, or resource.

  Rights include none, read, write, execute and all.  Use | to combine them,
  for example: "read | write" to permit read from, or write to, a path.

  Use a glob expression as a pattern.

  Suppose we do not want users to process MPEG video images:

    <policy domain="delegate" rights="none" pattern="mpeg:decode" />

  Here we do not want users reading images from HTTP:

    <policy domain="coder" rights="none" pattern="HTTP" />

  The /repository file system is restricted to read only.  We use a glob
  expression to match all paths that start with /repository:

    <policy domain="path" rights="read" pattern="/repository/*" />

  Lets prevent users from executing any image filters:

    <policy domain="filter" rights="none" pattern="*" />

  Any large image is cached to disk rather than memory:

    <policy domain="resource" name="area" value="1GP"/>

  Use the default system font unless overwridden by the application:

    <policy domain="system" name="font" value="/usr/share/fonts/favorite.ttf"/>

  Define arguments for the memory, map, area, width, height and disk resources
  with SI prefixes (.e.g 100MB).  In addition, resource policies are maximums
  for each instance of ImageMagick (e.g. policy memory limit 1GB, -limit 2GB
  exceeds policy maximum so memory limit is 1GB).

  Rules are processed in order.  Here we want to restrict ImageMagick to only
  read or write a small subset of proven web-safe image types:

    <policy domain="delegate" rights="none" pattern="*" />
    <policy domain="filter" rights="none" pattern="*" />
    <policy domain="coder" rights="none" pattern="*" />
    <policy domain="coder" rights="read|write" pattern="{GIF,JPEG,PNG,WEBP}" />
-->
<policymap>
  <!-- <policy domain="resource" name="temporary-path" value="/tmp"/> -->
  <policy domain="resource" name="memory" value="256MiB"/>
  <policy domain="resource" name="map" value="512MiB"/>
  <policy domain="resource" name="width" value="16KP"/>
  <policy domain="resource" name="height" value="16KP"/>
  <!-- <policy domain="resource" name="list-length" value="128"/> -->
  <policy domain="resource" name="area" value="128MP"/>
  <policy domain="resource" name="disk" value="1GiB"/>
  <!-- <policy domain="resource" name="file" value="768"/> -->
  <!-- <policy domain="resource" name="thread" value="4"/> -->
  <!-- <policy domain="resource" name="throttle" value="0"/> -->
  <!-- <policy domain="resource" name="time" value="3600"/> -->
  <!-- <policy domain="coder" rights="none" pattern="MVG" /> -->
  <!-- <policy domain="module" rights="none" pattern="{PS,PDF,XPS}" /> -->
  <!-- <policy domain="path" rights="none" pattern="@*" /> -->
  <!-- <policy domain="cache" name="memory-map" value="anonymous"/> -->
  <!-- <policy domain="cache" name="synchronize" value="True"/> -->
  <!-- <policy domain="cache" name="shared-secret" value="passphrase" stealth="true"/>
  <!-- <policy domain="system" name="max-memory-request" value="256MiB"/> -->
  <!-- <policy domain="system" name="shred" value="2"/> -->
  <!-- <policy domain="system" name="precision" value="6"/> -->
  <!-- <policy domain="system" name="font" value="/path/to/font.ttf"/> -->
  <!-- <policy domain="system" name="pixel-cache-memory" value="anonymous"/> -->
  <!-- <policy domain="system" name="shred" value="2"/> -->
  <!-- <policy domain="system" name="precision" value="6"/> -->
  <!-- not needed due to the need to use explicitly by mvg: -->
  <!-- <policy domain="delegate" rights="none" pattern="MVG" /> -->
  <!-- use curl -->
  <policy domain="delegate" rights="none" pattern="URL" />
  <policy domain="delegate" rights="none" pattern="HTTPS" />
  <policy domain="delegate" rights="none" pattern="HTTP" />
  <!-- in order to avoid to get image with password text -->
  <policy domain="path" rights="none" pattern="@*"/>
  <!-- disable ghostscript format types -->
  <policy domain="coder" rights="none" pattern="PS" />
  <policy domain="coder" rights="none" pattern="PS2" />
  <policy domain="coder" rights="none" pattern="PS3" />
  <policy domain="coder" rights="none" pattern="EPS" />
  <policy domain="coder" rights="none" pattern="PDF" />
  <policy domain="coder" rights="none" pattern="XPS" />
</policymap>
```
</details>

So there are differences, even ignoring (a) the commented lines, (b) the intentionally removed Ghostscript rules), such as the upstream `area` unit fix mentioned in #203.

I think we should include that fix from upstream, plus maybe the original doctype and add a comment explaining what our changes are?

For example (this is the original version with the commented lines removed,  the ghostscript rules removed, and an explanation of the Heroku changes):

```suggestion
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE policymap [
  <!ELEMENT policymap (policy)*>
  <!ATTLIST policymap xmlns CDATA #FIXED ''>
  <!ELEMENT policy EMPTY>
  <!ATTLIST policy xmlns CDATA #FIXED '' domain NMTOKEN #REQUIRED
    name NMTOKEN #IMPLIED pattern CDATA #IMPLIED rights NMTOKEN #IMPLIED
    stealth NMTOKEN #IMPLIED value CDATA #IMPLIED>
]>
<!--
  Configure ImageMagick policies.

  This is based on Ubuntu's default configuration with the following Heroku-specific changes:
  - The rules disabling the Ghostscript format types have been removed.
-->
<policymap>
  <policy domain="resource" name="memory" value="256MiB"/>
  <policy domain="resource" name="map" value="512MiB"/>
  <policy domain="resource" name="width" value="16KP"/>
  <policy domain="resource" name="height" value="16KP"/>
  <policy domain="resource" name="area" value="128MP"/>
  <policy domain="resource" name="disk" value="1GiB"/>
  <policy domain="delegate" rights="none" pattern="URL" />
  <policy domain="delegate" rights="none" pattern="HTTPS" />
  <policy domain="delegate" rights="none" pattern="HTTP" />
  <policy domain="path" rights="none" pattern="@*"/>
</policymap>
```

edmorley (approved):
This is great thank you!

PR #215, opened 2022-06-01
Publish images to private registry

joshwlewis:
Stack images need to be available for internal usage via a private, internal registry. This PR adds automatic image publication to an internal registry alongside the existing DockerHub publication process. 

The image tag naming constraints of the internal registry added some complexity to our naming convention. As a result, this PR also changes the behavior of `bin/build.sh` and `bin/publish-to-registries.sh`. Both of these now produce the entire family of images for a particular stack version, rather than just a runtime/build-time image pair.

Note that publishing to an internal registry is functionality already established in [pack-images](https://github.com/heroku/pack-images).

GUS-W-11227981
GUS-W-9864812


edmorley (approved):
Awesome!

PR #216, opened 2022-06-02
Add libyaml-dev to Heroku-22

schneems:
I was debugging a psych issue for getting the Ruby getting started guides working with Heroku-22 when I eventually realized that heroku-22 is missing libyaml-0-2. This is installed vial `libyaml-daev` which is present on the `heroku-22-build` but not `heroku-22`. This PR adds `libyaml-22`

Internal reference to the debugging thread: https://heroku.slack.com/archives/C02GZCPPV38/p1654192955601679


jonnymacs (approved):
👍 You mean `libyaml-dev` not `libyaml-daev` right?

joshwlewis (changes requested):
Did you mean to put `libyaml-dev` in the build image instead of the runtime image? This package is in heroku:20-build but not heroku:20. I think we should do the same here.

edmorley (changes requested):
@joshwlewis is correct, the `-dev` package should not be added to the runtime image.

We need to be extra careful about package changes in our stack images. A wrong step here can cause serious issues, either via build or runtime breakages, or else unnecessary packages in the wrong image, that we cannot later remove since it's then a breaking change. This should not be the kind of change we rubber stamp (there was a 3.5 minute gap between the ping in Slack and the approval granted here, which is not sufficient time to fully review the PR when not familiar with the packages in question), but take time to carefully review from SMEs of the stack images.

schneems:
Good catch and good call. 

I've updated the PR

edmorley (approved):
Thank you for the changes - looks great! :-)

I've confirmed that the `-dev` package is in place already in the build image variant:
https://github.com/heroku/stack-images/search?q=libyaml

joshwlewis (approved):

schneems:
Thanks for the review. I'm wondering what kind of reasonable but helpful automated checks we could add to help catch possible issues. Alternatively something lighter weight like a reviewer/contributor checklist? 

For example:

```
## Dev dependencies

List every dev dependency with an explanation of why it's needed. If modifying a runtime image (instead of build) state why the dependency needs to be available at runtime.
```

edmorley:
So I'm open to ideas, however I feel like that misses the main issues here, which were:
1. That reviews are not just an obstacle to deploying a change, but an important part of the process.
2. That some components/types of changes are riskier than others (like a change to stack-images).
3. The risk level is increased when both the author and reviewer of the change are less familiar with the component.
4. When the risk levels are high, it's best to ensure at least one of the people authoring or reviewing is more familiar with the component, or at least give enough time for others to chime in (should they choose to do so).

For changes to our language buildpacks, it's typical for the language owner to be the author of the PR (and so very familiar with the subject matter), in which case a less thorough review can be fine (and often unavoidable, given that others on the team won't fully understand the nuances of that language). However, I think for our shared components that doesn't apply as much, so we need to approach reviews of them with a less slightly less relaxed mindset :-)

edmorley:
I've triggered a new stack image release which includes this change:
https://github.com/heroku/stack-images/releases/tag/v80

PR #227, opened 2022-09-06
Added ARM64 support with multiarch build

### LINKED ISSUE #194: ARM64 support for image
Hello I'm trying to use this image on an M1 mac but I'm facing the same issue reported [here](https://github.com/docker/for-mac/issues/5321). The solution for this issue, which worked for me with other images is to use `arm64` compatible images.

After seeing the supported arch/os for the images I noticed only `amd64` is supported. Would it be possible to add support for `arm64`?

Thanks

boboldehampsink:
Closes #194

This is an attempt at creating a non-intrusive multiarch build. It adds an arm64 image without compromise.


boboldehampsink:
To really get this to work, the images should be pushed to registry directly by the build command

boboldehampsink:
As a proof of concept I pushed to a repo I owned, and you see that it has both architectures: https://hub.docker.com/repository/docker/robuust/heroku/tags

heroku-22/installed-packages.txt
```diff
@@ -306,7 +306,6 @@ mawk
 media-types
 mlock
 mount
-mtools
```
edmorley, line 309: Removing packages is a breaking change, and since Heroku apps are continuously rebased on newer stack images (so that they pick up security updates), any changes like that will not only break new builds moving forwards, but existing apps too (even those that haven't been built recently).

As such, this removal (and the syslinux change below) are not something we can accept for existing stacks.

Options are to:
1. Try and find a way to still include this package on all architectures
2. (If the package just doesn't exist for ARM64) Keep the package for x86_64, and omit on ARM64 only. (Though this then means there's even less parity between images, meaning yet another dev-prod parity difference.)
3. Wait until Heroku-24, where the package can be removed from all architectures

edmorley (changes requested):
Hi!

Thank you for the PR.

I wasn't going to leave a review until the PR was taken out of draft status (or you asked for feedback specifically), however judging by the comments in #194 people seem to (understandably, but unfortunately mistakenly) think this is ready - so leaving this to reduce confusion.

Before this could be merged the following would be needed:
1. Addressing the review comments below, and also adding docs to this repo.
2. Figuring out/testing the publishing part in CI (my understanding of `buildx` is that the pushing has to be done at the same time as building via the `--push` command, but that will need a rethink/refactor of the way builds/releasing via CI is performed)
3. (The biggest part) Adding support to all official buildpacks (and ideally most popular non-official buildpacks) for ARM64, otherwise all ARM64 users will be silently broken as soon as this is released. (Or at the very least, adding errors saying "this buildpack doesn't yet support ARM64, please force use of x86_64 when using this buildpack: <docker `--platform` instructions>" etc.)
4. Figuring out comms plan for rolling it out (eg Heroku changelog post) to reduce the end user impact/confusion from the long tail of other buildpacks that will not ever support ARM64.

I obviously don't expect you to do any of that work (it's our responsibility to do it), but this PR can't be merged until that has occurred.

In the meantime I would recommend using the images from your own Docker Hub repository, and if there are specific buildpacks where ARM64 support would help you, opening issues against those (and if you were feeling particularly adventurous, opening PRs if you didn't want to wait for us to add support, which may still be a fair way out).

heroku-22/installed-packages.txt
dzuelke: This package seems to be available on arm64 on all recent Ubuntu versions, no?

https://packages.ubuntu.com/bionic/mtools

boboldehampsink: mtools is not explicitly included, but is a dependency of syslinux. It can be required explicitly for ARM64, yes. Syslinux however cannot. I did some research and it seems that package is almost specifically built for x86, so there is no way to include that package on all architectures. 

So, to make this work we have to go with either option 2 or 3 as @edmorley suggested. However, Heroku 22 is only just out, so waiting for Heroku 24 could be a while.

@edmorley do we actually need a modified installed-packages.txt for ARM64 tho? Where is that used, in Heroku itself only right? And Heroku doesn't use ARM64 servers right? My original idea was that if you'd build this Docker image in the original way, it would just build the x86 image the way it was before. Only when you'd use buildx it would also build for arm64, which could differ a bit, yes.

edmorley: > do we actually need a modified installed-packages.txt for ARM64 tho? Where is that used, in Heroku itself only right?

The file is effectively a test fixture, it's not used outside of this repository. It's there to avoid hidden package changes as upstream make changes, to avoid regressions like #71. (Some packages will occasionally change their transitive dependency tree, which is not a breaking change for that particular package, but is a breaking change for the stack-images as a whole, if that only-installed-transitively package was something end users were depending on.)

As such, yes we need this test for all architectures.

boboldehampsink: Is there an easy way to make APT skip a package without failing? So we can just require syslinux but it will omit it for ARM64? Otherwise we would have to create something like a multi-stage docker build perhaps?

Also, `write_package_list` can be done for different architectures by referencing the SHA265 hash - see [Examples](https://docs.docker.com/build/building/multi-platform/#example) here

edmorley: A multi-stage build is not what we need here. Bear in mind that the x86_64 and arm64 Docker images will be built separately, whereas multi-stage builds refer to a single build that is then performed in several stages.

The solution needed is:

> Doing this would require adding architecture conditionals to build.sh

ie: Adding bash conditionals that check the output of `uname -m`.

boboldehampsink: This is actually possible with buildx:

```bash
if [ "$TARGETARCH" == "amd64" ]; then apt-get install -y --no-install-recommends syslinux; fi
```

bin/publish-to-registries.sh
```diff
@@ -38,10 +37,4 @@ if [[ -v CIRCLE_TAG ]]; then
 
   # Push release tags to internal registry
   push_group "${internalTag}" ".${CIRCLE_TAG}"
-
-  # Push latest/no-suffix tags to dockerhub (e.g. heroku/heroku:22)
-  push_group "${publicTag}" ""
-
-  # Push latest/no-suffix tags to internal registry
-  push_group "${internalTag}" ""
```
boboldehampsink, line 46: This still needs to be accounted for when pushing inline with buildx

heroku-22/installed-packages.txt
boboldehampsink: I have updated the PR with this logic

boboldehampsink:
I have added pushing with buildx

bin/publish-to-registries.sh
edmorley: So I think one of the issues here, is that Docker doesn't support having multiple architectures represented as one tag locally. So if you retag a local tag with an alias, then push that alias, I don't think it will push the special multi-architecture manifest, only the single architecture of that tag.

Instead, I think use of the `manifest` command will be required:
https://docs.docker.com/engine/reference/commandline/manifest/

See also:
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

edmorley: Ah yes, even better :-)

boboldehampsink: Added that (untested in your ci)

boboldehampsink:
@edmorley I will leave this here as an example, but I think the actual implementation of this is better done by you guys at Heroku. Let's say this is a proof of concept.

ShamoX:
@boboldehampsink, is your last comment meaning that this will not be merged by Heroku team?

If that so, is there any plan for integrating this feature?

I cannot do things locally, since my goal would be to get Gitlab CI/CD workflow (AutoDevOps one) works on my Runners, which are dispatched between my Kubernetes AMD64 and my macs (M1 and M2).

boboldehampsink:
@ShamoX see @edmorley's latest comment in https://github.com/heroku/stack-images/issues/194

joshwlewis:
@boboldehampsink Thanks a bunch for the inspiration and proof of concept. We just merged #245, which added heroku/heroku:24.nightly as a multi-arch image with amd/arm support. I'm going to close this, since we're not considering multi-arch or arm for heroku-22 and prior. Note that the upstream ubuntu 24.04 isn't stabilized or released yet, so heroku/heroku:24.nightly is likely to have substantial churn for a bit.

PR #228, opened 2022-09-19
Switch from Circle CI to GitHub Actions

mgomes:
Pipeline migrated from [Circle CI](https://app.circleci.com/pipelines/github/heroku/stack-images) :tada:

## Manual steps

Perform the follow steps to complete the migration:

### heroku/stack-images/build
- [x] Ensure environment variable is updated: `DOCKER_HUB_TOKEN`
- [x] Ensure environment variable is updated: `DOCKER_HUB_USERNAME`
- [x] Ensure environment variable is updated: `ID_SERVICE_PASSWORD`
- [x] Ensure environment variable is updated: `ID_SERVICE_TOKEN_ENDPOINT`
- [x] Ensure environment variable is updated: `ID_SERVICE_USERNAME`
- [x] Ensure environment variable is updated: `INTERNAL_REGISTRY_HOST`
- [x] Ensure environment variable is updated: `INTERNAL_REGISTRY_USERNAME`
- [x] Ensure environment variable is updated: `MANIFEST_APP_TOKEN`
- [x] Ensure environment variable is updated: `MANIFEST_APP_URL`

GUS-W-11783731.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+        status="$(git status --porcelain)"
+        if [[ -n "$status" ]]; then
+          echo "Generated files differ from checked-in versions! Run bin/build.sh to regenerate them locally."
+          echo -e "\nChanged files:\n${status}\n"
+          git diff
+          exit 1
+        fi
+    - name: Publish to image registries
+      run: bin/publish-to-registries.sh
+      if: "'main' == ${{ github.ref }} || ${{ github.ref }}"
```
edmorley, line 55: The publishing step is occurring on this PR branch at the moment, rather than just main (it didn't succeed, since I intentionally left the creds unset in the repo settings, just in case). 

The builder repo does this for only running on main:
https://github.com/heroku/builder/blob/01659d4b7bc72eb923c45a3c08033b27dad9a42a/.github/workflows/build-test-publish.yml#L122

Though this repo also needs this step to run on tags too, so it will need tweaking :-)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+    - name: Convert docker image and release to Heroku staging
+      run: bin/convert-and-publish-to-heroku.sh
+      if: "${{ github.ref }}"
```
edmorley, line 58: This step needs to run only on tag (not even on main)

edmorley (commented):
These are the env vars that we mentioned in the meeting the other month that will need updating to GitHub action equivalents:
https://github.com/heroku/stack-images/search?q=CIRCLE_TAG
https://github.com/heroku/stack-images/search?q=CIRCLE_SHA1

.github/workflows/ci.yml
jonnymacs: Thanks Ed. I will work on these changes.

edmorley (commented):
Ah one other thing I just remembered - Dependabot can be enabled for keeping the Actions dependencies in the GitHub Actions workflow up to date (Dependabot doesn't support doing this for Circle CI). Example config:
https://github.com/heroku/heroku-buildpack-python/blob/26af2d92a4f69bcc6b6db630c6237e4a3cd4f161/.github/dependabot.yml#L11-L14

.github/workflows/ci.yml
jonnymacs: I used the github.ref_type to check the ref is a tag, and the github.ref_name to assert it's main

.github/workflows/ci.yml
jonnymacs: I used the github.ref_type to check the ref is a tag

I also set an env var with the tag name from the ref_name. This allows for the logic in capture-docker-stack to remain dependent on the this value being set.

jonnymacs:
> These are the env vars that we mentioned in the meeting the other month that will need updating to GitHub action equivalents: https://github.com/heroku/stack-images/search?q=CIRCLE_TAG https://github.com/heroku/stack-images/search?q=CIRCLE_SHA1

GITHUB_SHA is a thing with github actions. GITHUB_TAG is not. Instead of checking if the build is a tag in this script, we depend on a new env var called GITHUB_TAG being set.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+          exit 1
+        fi
+    - name: Publish to image registries
+      run: bin/publish-to-registries.sh
+      if: success() && (github.ref_name == 'main' || github.ref_type == 'tag')
+    - name: Convert docker image and release to Heroku staging
+      run: bin/convert-and-publish-to-heroku.sh
+      if: success() && github.ref_type == 'tag'
+      env:
+        GITHUB_TAG: ${{ github.ref_name }}
```
edmorley, line 60: (optional)

Initially I thought there was a bug here, since this step passes `GITHUB_TAG`, but yet the step before (which also needs to run on tags) does not pass `GITHUB_TAG` (ie it seems the step prior should too).

However, `bin/publish-to-registries.sh` uses a different approach - it has a `"$GITHUB_REF_TYPE" == 'tag'` conditional, and then uses `GITHUB_REF_NAME` (which is equal to the tag name, so long as the first conditional is true).

I wonder if having these two different approaches might be confusing to future maintainers? In which case, maybe it's worth either:
1. Switching both to `GITHUB_TAG` (though that would need to not set the env var when `github.ref_name == 'main'` is the case),
2. Or else, removing the concept of `GITHUB_TAG` and adding `GITHUB_REF_TYPE" == 'tag'` +  `GITHUB_REF_NAME` to `tools/bin/capture-docker-stack`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+  build-stack:
+    runs-on: ubuntu-22.04
+    needs:
+    - shellcheck
+    env:
+      STACK: heroku-${{ matrix.stack-version }}
+      STACK_VERSION: "${{ matrix.stack-version }}"
+      DOCKER_HUB_TOKEN: "${{ secrets.DOCKER_HUB_TOKEN }}"
+      ID_SERVICE_PASSWORD: "${{ secrets.ID_SERVICE_PASSWORD }}"
+      MANIFEST_APP_TOKEN: "${{ secrets.MANIFEST_APP_TOKEN }}"
```
edmorley, line 33: Some env vars were missing:

```suggestion
      DOCKER_HUB_TOKEN: "${{ secrets.DOCKER_HUB_TOKEN }}"
      DOCKER_HUB_USERNAME: "${{ secrets.DOCKER_HUB_USERNAME }}"
      ID_SERVICE_PASSWORD: "${{ secrets.ID_SERVICE_PASSWORD }}"
      ID_SERVICE_TOKEN_ENDPOINT: "${{ secrets.ID_SERVICE_TOKEN_ENDPOINT }}"
      ID_SERVICE_USERNAME: "${{ secrets.ID_SERVICE_USERNAME }}"
      INTERNAL_REGISTRY_HOST: "${{ secrets.INTERNAL_REGISTRY_HOST }}"
      INTERNAL_REGISTRY_USERNAME: "${{ secrets.INTERNAL_REGISTRY_USERNAME }}"
      MANIFEST_APP_TOKEN: "${{ secrets.MANIFEST_APP_TOKEN }}"
      MANIFEST_APP_URL: "${{ secrets.MANIFEST_APP_URL }}"
```

edmorley (changes requested):
Thank you for doing this!

edmorley:
I've set the remaining env vars on:
https://github.com/heroku/stack-images/settings/secrets/actions

edmorley (approved):
This is good to merge at your convenience, if you are happy with the changes I added in:
- d8a071fde93bd1775566cfd48dce96599c18038e
- d75a24d6691d2cc611df76a102049b02febca75f
- d60d47adf5cf62b4d44fabf55cdd1f1eb431f673

(I pushed them directly since some of the changes couldn't be made through the suggestions interface/were going to conflict with the ones that were)

.github/workflows/ci.yml
jonnymacs: yeah, so the previous logic in `bin/publish-to-registries` was conditional on the `CIRCLE_TAG` which is only set if the ref type is a tag. Not knowing if / how / when that script will be run, and in the interest of limiting setting custom env variables I thought using the ref_type was the correct flag.

I don't think it will be any more confusing for future maintainers than it was, in fact I think it might be less confusing because it's a more explicit check on the ref_type.

So I'd prefer option #2 here and update the logic in `capture-docker-stack` to check the same. Let me do that.

jonnymacs:
@edmorley let me know what you think of this last change 9104c99

tools/bin/capture-docker-stack
```diff
@@ -12,7 +12,7 @@ STACK_VERSION=$(echo "${STACK}" | cut -d '-' -f 2-)
 DOCKER_IMAGE=heroku/$STACK_NAME:$STACK_VERSION
 DOCKER_IMAGE_VERSION=$(docker inspect "${DOCKER_IMAGE}" | jq .[].Id | cut -d ':' -f 2 | cut -b 1-12)
 
-GIT_REF=${CIRCLE_TAG:-${CIRCLE_SHA1:?'Error: Either CIRCLE_TAG or CIRCLE_SHA1 must be set!'}}
+[[ ${GITHUB_REF_TYPE} == 'tag' ]] && GIT_REF=${GITHUB_REF_NAME:?'Error: GITHUB_REF_NAME must be set!'} || GIT_REF=${GITHUB_SHA:?'Error: either GITHUB_REF_NAME or GITHUB_SHA must be set!'}
```
edmorley, line 15: What do you think about using a multi-line conditional instead, to make it more readable?

Also, the SHA mode was something I added as part of adding TPS support, but we have never used it (and should probably not do an untagged release), so I think we could (/should) just remove the SHA support?

With that, perhaps this then becomes something like this?

```bash
if [[ ${GITHUB_REF_TYPE} == 'tag' ]]; then
    GIT_REF=${GITHUB_REF_NAME:?'Error: GITHUB_REF_NAME must be set!'}
else
    abort "Error: Attempting to publish to Heroku without having Git tagged!"
fi
```

edmorley: As an added bonus this protects against the GitHub Actions workflow config having a broken `if` conditional, and running publish by mistake.

edmorley:
The publish from main to the nightly tags worked well after merge:
https://github.com/heroku/stack-images/actions/runs/3363091672/jobs/5575733404#step:5:136

That just leaves the Heroku stack upload part that we'll be able to confirm works during tomorrow's stack image release.

edmorley:
I've updated the playbook in:
https://github.com/heroku/cheverny/pull/263

Plus removed the old env vars from Circle CI.

PR #229, opened 2022-10-11
Make registry publishing release step easier to debug

schneems:
- Add before and after log statements for the two login commands. This helps isolate when one of them fails versus previously this is the output:

```
+ set +x
Error: Cannot perform an interactive login from a non TTY device
```

- Change curl `-s` to `-sS` so that errors are output on failure
- Add `-e` to `jq` so that parse failures error out and stop the propagation to the next pipe

GUS-W-11917558.

### DIFF
--- a/bin/publish-to-registries.sh
+++ b/bin/publish-to-registries.sh
@@ -3,17 +3,20 @@
 set -euo pipefail
 set -x
 
-
 bin/build.sh "${STACK_VERSION}"
 
-# Disable tracing temporarily (with `set +x`) to prevent logging registry tokens.
-echo "Logging into Docker hub..."
-(set +x; echo "${DOCKER_HUB_TOKEN}" | docker login -u "${DOCKER_HUB_USERNAME}" --password-stdin)
-echo "Done"
+(
+  # Disable tracing (until the end of this subshell) to prevent logging registry tokens.
+  set +x
 
-echo "Logging into internal registry..."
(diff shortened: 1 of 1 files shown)


bin/publish-to-registries.sh
```diff
@@ -6,9 +6,14 @@ set -x
 
 bin/build.sh "${STACK_VERSION}"
 
-# Disable tracing temporarily to prevent logging registry tokens.
+# Disable tracing temporarily (with `set +x`) to prevent logging registry tokens.
+echo "Logging into Docker hub..."
```
edmorley, line 10: The `set -x` from above will mean both the echo command as typed, and the output from the echo command will be shown in the logs (ie: the message will look as though it's been repeated twice). To avoid that, the logging can be moved inside the subshell after the `set +x`.

Re the `echo "Done"`, the docker login command already outputs "Login Succeeded" in the successful case, so I think the "done" message can be omitted.

I think it might also be worth using a multi-line statement inside the subshell now we're adding more things in there, to make it more readable. We can even collapse both subshells into one.

So how about something like this:

```bash
(
    # Disable tracing (until the end of this subshell) to prevent logging registry tokens.
    set +x

    echo "Logging into Docker Hub..."
    echo "${DOCKER_HUB_TOKEN}" | docker login -u "${DOCKER_HUB_USERNAME}" --password-stdin

    echo "Logging into internal container registry..."
    curl -sSf --retry 3 -X POST "$ID_SERVICE_TOKEN_ENDPOINT" -d "{\"username\":\"${ID_SERVICE_USERNAME}\",\"password\":\"${ID_SERVICE_PASSWORD}\"}" \
        | jq -er ".raw_id_token" \
        | docker login "$INTERNAL_REGISTRY_HOST" -u "$INTERNAL_REGISTRY_USERNAME" --password-stdin
)
```

edmorley (changes requested):

bin/publish-to-registries.sh
edmorley: To speed up this PR, I've made this change myself in f963fc5.

PR #232, opened 2023-02-03
Bump postgresql libraries and packages to version 15

kbarber:
Release notes:
https://www.postgresql.org/docs/15/release-15.html

GUS-W-12353294.

Based on prior art here: https://github.com/heroku/stack-images/pull/206

### DIFF
--- a/heroku-18-build/installed-packages.txt
+++ b/heroku-18-build/installed-packages.txt
@@ -517,10 +517,10 @@ perl-modules-5.26
 pinentry-curses
 pkg-config
 poppler-data
-postgresql-client-14
+postgresql-client-15
 postgresql-client-common
 postgresql-common
-postgresql-server-dev-14
+postgresql-server-dev-15
 procps
 python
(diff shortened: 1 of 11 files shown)


edmorley (approved):
Checking the PG 15 release notes (I added a link to them in the PR description) there don't appear to be any breaking changes to `libpq` or `psql` other than this for psql:

> * Limit support of psql's backslash commands to servers running PostgreSQL 9.2 or later (Tom Lane)
>
>   Remove code that was only used when running with an older server. Commands that do not require any version-specific adjustments compared to 9.2 will still work.

...which seems fine given it's for such as old version of PostresSQL (that is no longer supported by Heroku Postgres or AWS RDS), and only backslash commands.

If don't hear any concerns from the rest of the languages team about the above on Monday, I will merge this PR and the change will be released as part of the coming week's stack image update (GUS-W-12356656).

edmorley:
Thank you for the PR :-)

PR #238, opened 2023-03-06
maint: Add AWS RDS CA to trust store

mble-sfdc:
We've had a few requests on better supporting folks running their databases on AWS RDS or AWS Aurora. In order to support these folks without additional buildpacks etc, lets add the CA certs into the trust store.

We grab them directly from AWS, then check the SHA matches from the last time this was updated - this will infrequently change. We then split these certificates into individual certs so `update-ca-certificates` can support loading them.

Ref: GUS-W-12643488

### DIFF
--- a/heroku-18/setup.sh
+++ b/heroku-18/setup.sh
@@ -259,6 +259,17 @@ apt-get purge -y openjdk-8-jre-headless
 apt-get autoremove -y --purge
 test "$(file -b /etc/ssl/certs/java/cacerts)" = "Java KeyStore"
 
+# Install AWS RDS global CA bundle (https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.CertificatesAllRegions)
+mkdir -p /usr/local/share/ca-certificates/rds-ca-certs
+pushd /usr/local/share/ca-certificates/rds-ca-certs && \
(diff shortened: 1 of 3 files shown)


mble-sfdc:
Draft until I confirm a couple of things.

dzuelke:
I think you can simply do `curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem > /usr/local/share/ca-certificates/aws-rds-aurora.crt` and then `update-ca-certificates`.

Malax:
Another thing to keep in mind before we merge this: we need to make sure the Java `cacerts` will also contain these certs for consistency. `update-ca-certificates` might do this already.

dzuelke:
Yup I was also looking at that, @Malax. I believe these AWS certs need to be installed before the Java certs.

mble-sfdc:
Ack, sounds good. I don't have an RDS/Aurora handy to test with at the moment, hence the draft. Once I do have one, I'll be able to make sure everything is in the right place for different clients, including the Java trust store.

Malax:
@dzuelke thanks for checking this! 👍🏻

mble-sfdc:
Coming back to this (eventually) I am going to revisit this a little bit in terms of structure. I need to break up the certs for the Java keystore, but for ease of use with `libpq` based things I need to provide the single bundle in an accessible location.

I'll noodle on this then ask for review when ready.

mble-sfdc:
I'm punting on this for a while, but will revisit later.

PR #247, opened 2024-02-12
Stop publishing to the internal container registry

edmorley:
Currently the base images are published not only to Docker Hub, but also an internal container registry.

However, the last remaining consumer of the base-images from the internal registry is migrating away from these images as of:
https://github.com/heroku/kodon/pull/483

As such, there is no longer any need to publish to this registry, which simplifies the publishing process here, and means one less set of credentials to manage/rotate.

GUS-W-15019312.


Malax (approved):

edmorley:
I've deleted all `ID_SERVICE_*` and `INTERNAL_REGISTRY_*` secrets from:
https://github.com/heroku/base-images/settings/secrets/actions

PR #251, opened 2024-03-01
Add CNB_TARGET environment variables

joshwlewis:
`CNB_TARGET_*` environment variables are provided by `lifecycle` for CNB Platform API > 0.12 ([here](https://github.com/buildpacks/lifecycle/blob/be881f29d5b4a33795e1b61bb262ce9ae439cc10/phase/analyzer.go#L84-L93)), and [CNB Buildpack API 0.10 requires these  variables to be provided](https://github.com/buildpacks/spec/blob/buildpack/0.10/buildpack.md?plain=1#L750-L760).

However, when a build is using a Platform API < 0.12, but a Buildpack API >= 0.10, `CNB_TARGET_OS` and `CNB_TARGET_ARCH` are not provided by lifecycle. Which means buildpacks built against Buildpack API 0.10 do not see the env vars the spec says are mandatory. [`libcnb.rs`](https://github.com/heroku/libcnb.rs) versions 0.18.0 and newer use Buildpack API 0.10, and therefore expect these env vars to exist, and fail builds as a result.

This introduces `CNB_TARGET_OS` and `CNB_TARGET_ARCH` to the build images, so buildpacks using Buildpack API 0.10 (like any buildpack using libcnb.rs 0.18.0 or better) still have access to these environment variables even in scenarios where the CNB Platform API is < 0.12.

These variables could be removed at a later date, once https://github.com/buildpacks/lifecycle/issues/1308 is resolved.

Thread about this issue: https://salesforce-internal.slack.com/archives/C01R6FJ738U/p1709318630497849


heroku-24-build/Dockerfile
```diff
@@ -8,3 +8,5 @@ USER 1002
 ENV CNB_USER_ID=1002
 ENV CNB_GROUP_ID=1000
 ENV CNB_STACK_ID "heroku-24"
+ENV CNB_TARGET_OS="linux"
+ENV CNB_TARGET_ARCH="amd64"
```
joshwlewis, line 12: Obviously, this is not correct for arm builds. This probably needs to be DOCKERFILE arg.

ryanbrainard: Do we want to make this change now and then parameterize or leave it alone for now? I haven't looked how the build process is for this one.

ryanbrainard (approved):
LGTM with caveat on heroku-24-build

heroku-20-cnb-build/Dockerfile
```diff
@@ -9,8 +9,10 @@ RUN mkdir /app && \
 
 ENV CNB_USER_ID=1000
 ENV CNB_GROUP_ID=1000
-
 ENV CNB_STACK_ID "heroku-20"
+ENV CNB_TARGET_OS="linux"
```
edmorley, line 13: For consistency, would it make sense to use `$TARGETOS` etc for these?

Or are they only available when using buildx?

That said, is there a reason why heroku-20/22 aren't using buildx too?
https://github.com/heroku/base-images/blob/d1ab0cb6a28e67c5a42f95e86c661dee1523768c/bin/build.sh#L54-L69

...I would guess to reduce number of changes to existing stacks?  But longer term should we consolidate and use buildx for the older stacks too (albeit in single arch mode)?

heroku-24-build/Dockerfile
```diff
@@ -8,3 +8,5 @@ USER 1002
 ENV CNB_USER_ID=1002
 ENV CNB_GROUP_ID=1000
 ENV CNB_STACK_ID "heroku-24"
+ENV CNB_TARGET_OS=$TARGETOS
+ENV CNB_TARGET_ARCH=$TARGETARCH
```
edmorley, line 12: Once the upstream `lifecycle` bug is fixed and our builder updated to that version, will we remove these? I presume it will be safe to do so, since once `lifecycle` is setting them for us there won't be backwards compat concerns for removing an env var that was once set before?

heroku-20-cnb-build/Dockerfile
joshwlewis: > Or are they only available when using buildx?

Yes, that's my understanding. Read that here: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/.

> That said, is there a reason why heroku-20/22 aren't using buildx too? I would guess to reduce number of changes to existing stacks?

edmorley (approved):

PR #252, opened 2024-03-04
Create multi-arch manifest copy, rather than retagging.

joshwlewis:
`docker tag SOURCE TARGET` isn't aware of multi-arch image manifests. `docker tag` therefore isn't a copy of a manifest, but a new single-arch image. 

For multi-arch images / manifest lists, we should instead use `docker buildx imagetools create`, so the manifest list is preserved.


edmorley (approved):
Thank you :-)

PR #256, opened 2024-03-11
Use `RUN --mount` instead of `COPY` or inlining files

edmorley:
This backports the `RUN --mount` change from Heroku-24 to Heroku-20 and Heroku-22, which saves needing to `COPY` in the temporary `setup.sh` script, thereby:
(a) reducing layer count (by 1 for the runtime images and 2 for the build variants),
(b) avoiding the deleted `setup.sh` being shipped in the image (not accessible, since it's in an earlier layer).

In addition, this allows us to stop inlining the Postgres APT GPG key and ImageMagick policy in `setup.sh`, making the script easier to read.

I've left the inlined APT sources list, since:
(a) they use concatenation and so can't be replaced by a plain file copy,
(b) they are smaller to inline,
(c) the APT sources are arguably more important content to see upfront when reading `setup.sh`, without needing to switch to another file.


Malax (approved):
TIL about "RUN --mount" 👍🏻

PR #258, opened 2024-03-11
Cleanup setup scripts

edmorley:
* Removes the stderr redirection to stdout, since it's no longer necessary (we're no longer using Travis CI, where stderr was highlighted in bright red)
* Combines the `set` calls
* Switches to an array for the list of packages to install, which avoids the need for line continuation markers, and will allow us to add code comments later explaining why we install certain packages (comments break lines using continuation markers)


Malax (approved):
Love the change to arrays!

PR #260, opened 2024-03-11
Remove `CNB_TARGET_*` env vars workaround

edmorley:
These were added recently in #251 to work around this upstream `lifecycle` bug:
https://github.com/buildpacks/lifecycle/issues/1308

The fix for that has since been released in lifecycle 0.19.0:
https://github.com/buildpacks/lifecycle/releases/tag/v0.19.0

Which our builder images are now using:
https://github.com/heroku/cnb-builder-images/pull/475

As such, we no longer need the workaround here (and in fact, setting these env vars now cannot have an effect, since `lifecycle` now ignores them).


Malax (approved):

edmorley:
I might need to revisit this, pending:
https://github.com/buildpacks/lifecycle/pull/1309#issuecomment-1988929012

I also found a gap in test coverage, for which I opened:
https://github.com/heroku/cnb-builder-images/pull/480

edmorley:
> The fix for that has since been released in lifecycle 0.19.0:
> https://github.com/buildpacks/lifecycle/releases/tag/v0.19.0

And backported to v0.17.5:
https://github.com/buildpacks/lifecycle/releases/tag/v0.17.5

PR #261, opened 2024-03-13
Test `.img` file generation in CI on PRs and `main` too

edmorley:
The base images are built as Docker images and on release, uploaded to Docker Hub. However, the Heroku platform currently doesn't use those images directly, instead using a custom `.img` file from S3.

Previously this `.img` file generation was only performed on release, meaning that the CI for PRs and on `main` wouldn't test that code-path, so we might not find out about breakage until time of release.

Now, the generation (but not S3 upload) is performed for PRs / `main` too.

This will allow us to more easily iterate on optimising the image generation in later PRs.

GUS-W-15244492.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,5 @@ jobs:
       - name: Unpublish temp tags from this run
         run: bin/unpublish-tags.sh
         if: always()
-      - name: Convert docker image and release to Heroku staging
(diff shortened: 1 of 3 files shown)


edmorley:
See here for an example of the image generation being run, but the upload being skipped (as expected, since a PR):
https://github.com/heroku/base-images/actions/runs/8267747525/job/22618965845?pr=261#step:7:1

colincasey (approved):

PR #269, opened 2024-03-19
Upgrade Heroku-20 and Heroku-22 to PostgreSQL 16

brittanyrjones:
updating to major postgres version 16
pg16 Release notes: https://www.postgresql.org/docs/release/16.0/

GUS-W-15288386.

### DIFF
--- a/heroku-20-build/installed-packages.txt
+++ b/heroku-20-build/installed-packages.txt
@@ -561,10 +561,10 @@ pinentry-curses
 pkg-config
 poppler-data
 poppler-utils
-postgresql-client-15
+postgresql-client-16
 postgresql-client-common
 postgresql-common
-postgresql-server-dev-15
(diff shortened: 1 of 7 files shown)


edmorley:
Hi! Thank you for the PR. CI was failing due to an oversight in the GHA conditionals, which meant credentials were required for PRs (which aren't present on PRs from forks) - which I've fixed in #270, and rebased this PR onto.

edmorley (approved):
Thank you for the PR!

I've checked the release notes for anything relating to `libpq`, `psql` or mentioning the word "client" and there don't seem to be any breaking changes that will cause issues for existing apps, so this looks fine to me.

I also confirmed that we didn't need to update Heroku-24 too (which is a WIP), since it appears it's already using PostgreSQL 16.

Issue #266, opened 2024-03-18
Reduce size of Heroku-24 run and build images


edmorley:
Initial experimental (since Ubuntu 24.04 isn't even GA itself yet) Heroku-24 images were added in #245.

In that PR, a few packages were dropped compared to Heroku-22, to try and reduce the image size.

However, we'd like to reduce the size of the image further, since even with those changes the images have ended up larger than Heroku-22:

```
$ docker images
REPOSITORY      TAG        IMAGE ID       CREATED      SIZE
heroku/heroku   24-build   19fbe99496c9   7 days ago   1.09GB
heroku/heroku   24         260534c30d08   7 days ago   665MB
heroku/heroku   22-build   9bd3e3a84bac   7 days ago   1.02GB
heroku/heroku   22         37fee4e277ea   7 days ago   643MB
```

Smaller images sizes are going to be even more important in a CNB world, where the image size trade-offs have shifted quite a bit from the stack+slug model. In addition, in an SBOM world, reducing number of packages (and thus potential vulnerability surface area) is going to be something users become increasingly interested in.

We can't remove packages from a new base image version once it GAs (due to image rebasing meaning every new image update must be backwards compatible), so we must do this before Heroku-24 GAs.

Possible ideas:
- Remove build toolchains from the run image (they were added in #127 for Ruby MJIT support, but MJIT has now mostly been superseded by YJIT)
- Remove git from the run image
- Remove Python from the run image
- Remove stunnel (since it pulls in systemd too, and Heroku Redis now supports TLS natively, making the stunnel buildpack redundant)

GUS-W-15159536.

edmorley:
> Smaller images sizes are going to be even more important in a CNB world

As already being seen in:
https://github.com/heroku/buildpacks/discussions/6#discussioncomment-9306826
https://www.reddit.com/r/rails/comments/1chrwmq/comment/l24ylu2/

edmorley:
Ok, we're now in a much better place (and as good as we're going to get for now; longer term we can also discuss having a separate slim variant)...

Before:

```
$ docker images
REPOSITORY      TAG        IMAGE ID       CREATED      SIZE
heroku/heroku   24-build   19fbe99496c9   7 days ago   1.09GB
heroku/heroku   24         260534c30d08   7 days ago   665MB
```

After:

```
$ docker images
REPOSITORY      TAG        IMAGE ID       CREATED          SIZE
heroku/heroku   24-build   f324a55dc0cd   29 minutes ago   959MB
[...]

PR #273, opened 2024-03-20
Heroku-24: Remove gcc, make and libc6-dev from the run image

edmorley:
(This is a first step towards reducing image size for Heroku-24: #266)

GCC was added to our run images back in #127 in order to support Ruby 2.6's then new MJIT feature:
https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/

However, since then:
- The Ruby MJIT feature hasn't really resulted in significant performance benefits for real world use-cases like a Rails app.
- Ruby's MJIT has since been superseded by YJIT, which is faster and doesn't need GCC at runtime:
    - https://shopify.engineering/yjit-just-in-time-compiler-cruby
    - https://shopify.engineering/ruby-yjit-is-production-ready
- The image size impact of including build tools in our run images has increased considerably (#127 quoted it as 84 MB, but measuring now it's 203 MB).
- In a CNB world, image size is much more of a concern than in the S3 `.img` + slug model, so we need to be more selective over what packages we include.

As such, this removes `gcc`, `make` and `libc6-dev` from the run image for a 203 MB saving (they are still present in the build image, hence zero changes to `installed-packages-*.txt` for that image).

Richard (Ruby owner) has confirmed he's fine with this change.

Note: I'm intentionally not adding [`binutils`](https://packages.ubuntu.com/noble/amd64/binutils/filelist) back (which was a transitive dependency), since its 15 MB cost is not worth it for the ~once a year platform operator debugging use-case. (More discussion on this at:
https://salesforce-internal.slack.com/archives/C02GZCPPV38/p1709204898451809?thread_ts=1709054633.272659&cid=C02GZCPPV38)

Before:

```
-----> Size breakdown...
       heroku/heroku:24         661MB
       heroku/heroku:24-build   1.13GB
```

After:

```
-----> Size breakdown...
       heroku/heroku:24         458MB  (203 MB reduction)
       heroku/heroku:24-build   1.13GB (unchanged)
```

Towards #266.
GUS-W-15159536.


runesoerensen (approved):
Seems we might as well get this going so we can discover issues before GA!

PR #281, opened 2024-03-27
Heroku-24: Use the same user for the run and build images

### LINKED ISSUE #268: Heroku-24: Implications of the switch to separate Linux users for build vs run images
As part of the initial Heroku-24 PR, we switched to using separate Linux users for the run and build images:
https://github.com/heroku/base-images/pull/245#discussion_r1482159182

This change was made because the spec [says](https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#run-image):

> * The image config's `User` field is set to a user with a DIFFERENT user [†](https://github.com/buildpacks/spec/blob/platform/0.13/README.md#operating-system-conventions)UID/[‡](https://github.com/buildpacks/spec/blob/platform/0.13/README.md#operating-system-conventions)SID as the build image.

This upstream spec change was made as a result of [RFC-85](https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md), which proposed the change for security reasons. See also the discussion in https://github.com/buildpacks/rfcs/pull/146 and [the working group recording](https://youtu.be/Kd_GMqMHkOo?feature=shared&t=1095).

The implications of the change is that at runtime the `/app` (or `/workspace`) directory and `/layers` directories are read-only to the default run image user. Meaning that in order to perform writes at run-time, either:
1. The app must write to `/tmp` or `$HOME` instead
2. The image must instead be run with the user that performed the build (eg `docker run --user heroku-build`)
3. A buildpack (or inline buildpack) must `chmod` specific directories/files as group-writeable during the build (which allows writes, since both the `heroku` and `heroku-build` users are in the same group).

This is more secure, but will mean:
- Some buildpacks/apps will need changes to either switch to the `/tmp` and `$HOME` directories, or to add `chmod`ing. For example:
    - https://github.com/heroku/buildpacks-nodejs/issues/800
    - https://github.com/heroku/buildpacks-php/issues/89
    - https://github.com/heroku/cnb-builder-images/actions/runs/8326016245 
- When debugging, people will need to explicitly switch to a different user (either locally or on Heroku) or else rebuild their app when making debugging changes.

As such it seems like quite a substantial change to make, given that we don't have control over app code, and that some ecosystems heavily rely upon writing to the app directory (for example Ruby Rails apps writing to `<app>/tmp/`). We could try things like having the Ruby buildpack chmod `<app>/tmp/` if present, however, there is likely a long tail of other cases that will break, and the error messages are not the easiest for customers to figure out.

The spec only says `SHOULD`, so we could choose not to make this change. (And perhaps consider it for Heroku-26, or else make read-only images at runtime an opt-in platform feature in the future.)

However, some other platforms out there already use separate users at runtime (as seen in https://github.com/heroku/buildpacks-nodejs/issues/800), and so regardless of what we chose for Heroku, if we want our buildpacks to work elsewhere, we’ll need to make sure they do not write to `/layer` or the app directory. We could always test in CI that this works, even if we chose not to have separate user (and this read-only `/layer` and app dir) for the Heroku base images themselves.

@heroku/languages Thoughts?

GUS-W-15342842.

edmorley:
The upstream CNB spec recently changed to say that build and run images `SHOULD` use a separate Linux user for each image:
https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md
https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#run-image

As such, when adding Heroku-24 we decided to use separate users:
https://github.com/heroku/base-images/pull/245#discussion_r1482159182

However, we found this is likely to cause a number of compatibility issues with existing apps and parts of the ecosystem (see #268).

Whilst we can (and will) adjust our own buildpacks to do the right thing (not write to `/layers/` or the app source directory at runtime), it's going to be some time before existing apps/frameworks/... make similar changes. In addition, the failure modes are not easy for users to debug or solve (they will have to know that seeing access denied errors means needing to use `chmod` to make directories group writeable in an inline buildpack step or similar).

As such, we're deferring making this switch for now, and will revisit in the future (either for Heroku-26, or as an opt-in feature for Heroku-24), when the various third party language ecosystems are more ready for this.

We will still be in compliance with the spec, since it says `SHOULD` not `MUST`.

We will also add integration testing to our own CNBs to ensure that they operate correctly in environments that do run split build/run users.

As part of this change, I've also switched the `heroku` user's ID from `1001` back to `1000`, for consistency with the Heroku-20/22 CNB base images.

I've also switched back to the `USER <name>` syntax instead of `USER <id>`, since both are permitted by the OCI and CNB specs, and the former is (a) IMO more intuitive (eg for users needing to switch to `root` and back in their own `Dockerfile`), (b) matches what Heroku-20/22 do.

See also:
https://manpages.ubuntu.com/manpages/noble/en/man8/userdel.8.html
https://manpages.ubuntu.com/manpages/noble/en/man8/groupadd.8.html

Closes #268.
GUS-W-15342842.

### DIFF
--- a/heroku-20-cnb-build/Dockerfile
+++ b/heroku-20-cnb-build/Dockerfile
@@ -7,6 +7,7 @@ RUN groupadd heroku --gid 1000 \
   && chown heroku:heroku /app
 
 # https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#build-image
+USER heroku
 ENV CNB_USER_ID=1000
 ENV CNB_GROUP_ID=1000
 # Note: This image doesn't inherit from the CNB run image variant so we have
@@ -20,5 +21,3 @@ LABEL io.buildpacks.base.maintainer="Heroku"
 # https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#iobuildpacksstack-labels
 ENV CNB_STACK_ID="heroku-20"
 LABEL io.buildpacks.stack.id="heroku-20"
-
-USER heroku
--- a/heroku-22-cnb-build/Dockerfile
+++ b/heroku-22-cnb-build/Dockerfile
@@ -7,6 +7,7 @@ RUN groupadd heroku --gid 1000 \
   && chown heroku:heroku /app
 
 # https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#build-image
+USER heroku
 ENV CNB_USER_ID=1000
 ENV CNB_GROUP_ID=1000
 # Note: This image doesn't inherit from the CNB run image variant so we have
@@ -20,5 +21,3 @@ LABEL io.buildpacks.base.maintainer="Heroku"
 # https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#iobuildpacksstack-labels
 ENV CNB_STACK_ID="heroku-22"
 LABEL io.buildpacks.stack.id="heroku-22"
-
-USER heroku
--- a/heroku-24-build/Dockerfile
+++ b/heroku-24-build/Dockerfile
@@ -1,14 +1,14 @@
 ARG BASE_IMAGE=heroku/heroku:24
 FROM $BASE_IMAGE
 
-# We have to temporarily switch back to root, since the run image sets a non-root default USER.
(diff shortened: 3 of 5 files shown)


joshwlewis (approved):
Thanks for tackling this!

PR #285, opened 2024-04-22
Heroku-24: Omit transitive deps from list of packages to install

edmorley:
This cleans up the packages list in `setup.sh` so that it only includes top-level dependencies, and no longer redundantly specifies packages that are transitive dependencies of the main packages.  In addition, `coreutils` has been omitted since it's already in the base Ubuntu 24.04 image (since it's in the `essentials` group).

(In the past before we had the installed packages list validated by CI, these redundant transitive deps served as a way to ensure we didn't accidentally omit a package.)

This change is a no-op in terms of the final images (note how there is no change to the generated `installed-packages-*.txt` lists), but makes it easier to follow/audit the remaining packages in the list.

GUS-W-15159536.

### DIFF
--- a/heroku-24/setup.sh
+++ b/heroku-24/setup.sh
@@ -42,10 +42,8 @@ packages=(
   # For dig, host and nslookup.
   bind9-dnsutils
   bzip2
-  coreutils
   curl
   file
-  fontconfig
   geoip-database
   gettext-base
   gir1.2-harfbuzz-0.0
@@ -55,13 +53,9 @@ packages=(
   iproute2
   iputils-tracepath
   less
-  libaom3
   libargon2-1
   libass9
   libc-client2007e
-  libcairo2
-  libcurl4
-  libdatrie1
   libdav1d7
   libev4
   libevent-2.1-7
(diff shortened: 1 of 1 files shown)


heroku-24/setup.sh
```diff
@@ -161,15 +127,12 @@ packages=(
   postgresql-client-16
   rename
   rsync
-  shared-mime-info
   socat
-  tar
```
joshwlewis, line 166: I presume it's removed here because it's already in `essentials`. My opinion, `tar` is critical to our systems and should be listed. If canonical decided it shouldn't be in `essentials` in ubuntu:26.04, I'd still want it to be in `heroku-26`.

joshwlewis (changes requested):
I'd like to see `tar` and potentially `coreutils` having explicit dependencies. Maybe we don't have to remove `essentials` here if those were the only two?

heroku-24/setup.sh
edmorley: Yeah I was on the fence whether to remove the explicit entry for it. My reasoning was that when adding a new stack, we'd diff the `installed-packages-*.txt` for the old and new stacks and see the omission.

However, I'm fine with adding it back too.

Ref `coreutils`, it includes packages that would never be omitted from the base image, so they would never not include it IMO.

edmorley: > Ref `coreutils`, it includes packages that would never be omitted from the base image

For example, `env` and `cat`:
https://packages.ubuntu.com/noble/amd64/coreutils/filelist

joshwlewis: > My reasoning was that when adding a new stack, we'd diff the installed-packages-*.txt for the old and new stacks and see the omission.

That was part of my reservation I think. Yes, I did diff the installed packages during `heroku-24` development, but there's nothing in our process that makes that a mandatory part of the process for new versions. This is a manual thing that must be remembered.

joshwlewis (approved):

PR #297, opened 2024-04-29
Heroku-24: Add more comments to the list of packages to install

edmorley:
Previously there was no explanation as to why most of the packages being installed were included in the images.

This adds an initial round of comments based on a combination of:
- `ldd` scans of runtimes/deps of built getting-started guide apps (plus in the case of PHP, an app set up by David that has all PHP extensions enabled)
- searching through Git history/GitHub PRs for this repo
- GitHub code-search of other `heroku` repos, incl non-official-language Heroku buildpacks
- my knowledge of the Heroku platform

It's not fully complete, but is better than what we had before :-)

I've switched to putting the comment on the same line as the package, since it allows the package list to be sorted automatically when using a "sort lines alphabetically" feature of an editor.

GUS-W-15616780.

### DIFF
--- a/heroku-24-build/setup.sh
+++ b/heroku-24-build/setup.sh
@@ -8,10 +8,9 @@ packages=(
   autoconf
   automake
   bison
-  # Includes gcc, g++, make, patch, libc6-dev etc.
-  build-essential
+  build-essential # Includes gcc, g++, make, patch, libc6-dev etc.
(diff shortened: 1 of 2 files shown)


joshwlewis (approved):

dzuelke (approved):
Thanks!

PR #308, opened 2024-05-22
Add `jq` to the run image for all stacks

edmorley:
So that Heroku Exec's init script can use it instead of Python's `json` module for extracting keys out of the Heroku Exec service API JSON response.

JQ was already in the build image, so this is a no-op for the build images, and only increases the run image size by ~470KB.

See:
https://gus.lightning.force.com/lightning/r/0D5EE00001l4DXC0A2/view
https://packages.ubuntu.com/noble/jq

GUS-W-15832128.

### DIFF
--- a/heroku-20-build/setup.sh
+++ b/heroku-20-build/setup.sh
@@ -13,7 +13,6 @@ packages=(
   cmake
   gettext
   git
-  jq
   libacl1-dev
   libapt-pkg-dev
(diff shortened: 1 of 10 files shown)


joshwlewis (approved):
There is customer demand for this, too: https://elements.heroku.com/buildpacks/chrismytton/heroku-buildpack-jq is fairly popular.

PR #316, opened 2024-07-23
Heroku-22: remove memory limitations from imagemagick

david-caro:
The size and memory limitations are not useful as they don't check the system's limits, but are just hard-coded, see:
https://github.com/ImageMagick/ImageMagick/issues/396#issuecomment-319569255

And specially for containers, the limits are better set by the runtime.

It would be less troublesome if you could override the settings from this configuration file, but unfortunately you can't.

### DIFF
--- a/heroku-22/imagemagick-policy.xml
+++ b/heroku-22/imagemagick-policy.xml
@@ -1,10 +1,4 @@
 <policymap>
-  <policy domain="resource" name="memory" value="256MiB"/>
-  <policy domain="resource" name="map" value="512MiB"/>
-  <policy domain="resource" name="width" value="16KP"/>
-  <policy domain="resource" name="height" value="16KP"/>
-  <policy domain="resource" name="area" value="128MP"/>
-  <policy domain="resource" name="disk" value="1GiB"/>
   <policy domain="delegate" rights="none" pattern="URL" />
   <policy domain="delegate" rights="none" pattern="HTTPS" />
   <policy domain="delegate" rights="none" pattern="HTTP" />


edmorley (changes requested):
Hi! 

In general, once a Heroku stack version (such as Heroku-22) is released, we cannot make breaking changes to it. Whilst at first glance this might not seem like a breaking change (since it removes limits, rather than setting them), it's quite possible that there are apps that are relying on these current limits to protect against inputs that would cause the app to run out of memory and be killed by the runtime entirely. 

For example: An app that accepts user image inputs and processes them, where the user can either accidentally or maliciously submit complex payloads that could cause resource exhaustion. With suitable imagemagick limits the impact of this payload would be limited to the single user request. Without these limits, the entire server process could fail or be killed by the runtime for OOM, which would impact requests being served by other workers/threads in the same dyno. Removing these limits for existing apps without warning (and the chance for them to make architectural changes) would not be safe or go down well with users.

We also generally try to avoid customising Ubuntu packages, instead preferring to use the stock package from upstream. The custom imagemagick config used for Heroku-22 and older was a slight departure from that - which is why for Heroku-24 we removed the custom config and instead rely on the default file provided by Ubuntu's `imagemagick-6-common` package: https://github.com/heroku/base-images/pull/245/commits/2806af80675d3e010e3921456efffcf9d17dc3dc

This default `imagemagick-6-common` Ubuntu 24.04 config file used in Heroku-24 has higher resource limits than the custom file we used before:

Heroku-22:
```
$ docker run --rm -it heroku/heroku:22 bash -c 'cat /etc/ImageMagick-6/policy.xml | grep "<policymap>" -A 999 | grep -v "<!--" | grep resource | sort'
  <policy domain="resource" name="area" value="128MP"/>
  <policy domain="resource" name="disk" value="1GiB"/>
  <policy domain="resource" name="height" value="16KP"/>
  <policy domain="resource" name="map" value="512MiB"/>
  <policy domain="resource" name="memory" value="256MiB"/>
  <policy domain="resource" name="width" value="16KP"/>
```

Heroku-24:
```
$ docker run --rm -it heroku/heroku:24 bash -c 'cat /etc/ImageMagick-6/policy.xml | grep "<policymap>" -A 999 | grep -v "<!--" | grep resource | sort'
  <policy domain="resource" name="area" value="256MP"/>
  <policy domain="resource" name="disk" value="2GiB"/>
  <policy domain="resource" name="height" value="32KP"/>
  <policy domain="resource" name="map" value="2048MiB"/>
  <policy domain="resource" name="memory" value="1024MiB"/>
  <policy domain="resource" name="width" value="32KP"/>
```

Hopefully these higher limits solve the issue you are seeing, given they are 2-4x higher than before, and so upgrading your app's stack to Heroku-24 will help.

If they do not, you will want to follow the steps here for overriding the policy:
https://help.heroku.com/RFDJQSG3/how-can-i-override-imagemagick-settings-in-a-policy-xml-file

If you would like to see changes to the default policy, then they would need to be made upstream in the Ubuntu `imagemagick-6-common` package:
https://launchpad.net/ubuntu/+source/imagemagick/+bugs

david-caro:
Thanks for the quick reply!

We are working on upstream debian to remove the sizes from the file (following up on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=929826), might take a look at ubuntu too.

We will probably have to override the policy with the link you passed (very helpful!).

I'm closing this for now, thanks again

PR #317, opened 2024-07-31
Add missing runtime libs to heroku-22 and heroku-24

dzuelke:
As a general rule, we make dynamic libraries available in the run base images if the packages containing headers for dynamic linking (typically named `lib…-dev`) are present in the build image.

This ensures that programs that dynamically link against libraries using these headers at build time (e.g. native extensions for various languages) work reliably at runtime.

A recent review of the list of packages on all stacks turned up a number of packages that aren't available at runtime, but have build time headers.

Ignoring `heroku-20` (as it is deprecated), and disregarding non-8-bit `libpcre…` libraries (which already are in the run images) as well as X11-related libraries, we're missing the following:

`heroku-22`:
- `libgeoip1` because of `libgeoip-dev`
- `libhashkit2` because of `libhashkit-dev`
- `libmemcachedutil2` because of `libmemcached-dev`
- `libnetpbm10` because of `libnetpbm10-dev`
- `libwmf-0.2-7` because of `libwmf-dev`

`heroku-24`:
- `libgirepository-1.0-1` because of `libgirepository1.0-dev` and `libgirepository-1.0-dev`
- `libidn12` because of `libidn-dev`
- `libmemcachedutil2t64` because of `libmemcached-dev`
- `libwebpdecoder3` because of `libwebp-dev`
- `libwmf-0.2-7` because of `libwmf-dev`

GUS-W-16381871

### DIFF
--- a/heroku-22/setup.sh
+++ b/heroku-22/setup.sh
@@ -73,6 +73,7 @@ packages=(
   libgd3
   libgdk-pixbuf2.0-0
   libgdk-pixbuf2.0-common
+  libgeoip1
   libgnutls-openssl27
   libgnutls30
   libgnutlsxx28
@@ -82,13 +83,16 @@ packages=(
   libharfbuzz-gobject0
   libharfbuzz-icu0
   libharfbuzz0b
+  libhashkit2
   libheif1
   liblzf1
   libmagickcore-6.q16-3-extra
   libmcrypt4
   libmemcached11
+  libmemcachedutil2
   libmp3lame0
   libmysqlclient21
+  libnetpbm10
   libnuma1
   libogg0
   libonig5
@@ -120,6 +124,7 @@ packages=(
   libwebp7
   libwebpdemux2
   libwebpmux3
+  libwmf-0.2-7
   libx264-163
   libx265-199
   libxcb-render0
--- a/heroku-24/setup.sh
+++ b/heroku-24/setup.sh
@@ -65,12 +65,15 @@ packages=(
   libevent-pthreads-2.1-7
   libgd3
   libgdk-pixbuf-2.0-0
+  libgirepository-1.0-1 # Because libgirepository1.0-dev and libgirepository-1.0-dev are in the build image
   libgnutls-openssl27
   libgnutls30 # Used by the Ruby and PHP runtimes.
   libharfbuzz-icu0 # Used by FFmpeg in heroku-buildpack-activestorage-preview.
+  libidn12 # Because libidn-dev is in the build image
   liblzf1 # Used by the PHP Redis extension.
   libmagickcore-6.q16-7-extra # Used by the PHP Imagick extension (using the `-extra` package for SVG support).
   libmemcached11 # Used by the PHP Memcached extension.
+  libmemcachedutil2t64 # Same -dev headers as libmemcached11
   libmp3lame0 # Used by FFmpeg in heroku-buildpack-activestorage-preview.
   libmysqlclient21
   libncurses6 # Used by the Ruby runtime.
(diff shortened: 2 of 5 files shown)


dzuelke:
Sizes:

```ShellSession
% docker run -u root --rm -ti heroku/heroku:22 bash
root@2e3c355b3af4:/# apt-get update
…
root@2e3c355b3af4:/# apt-get install libgeoip1 libhashkit2 libmemcachedutil2 libnetpbm10 libwmf-0.2-7
Reading package lists... Done
Building dependency tree... Done
Suggested packages:
  geoip-bin libwmf-0.2-7-gtk
  libgeoip1 libhashkit2 libmemcachedutil2 libnetpbm10 libwmf-0.2-7
0 upgraded, 5 newly installed, 0 to remove and 3 not upgraded.
Need to get 284 kB of archives.
After this operation, 1001 kB of additional disk space will be used.
…
```

```ShellSession
% docker run -u root --rm -ti heroku/heroku:24 bash      
root@5e5722257542:/# apt-get update
…
root@5e5722257542:/# apt-get install libgirepository-1.0-1 libidn12 libmemcachedutil2t64 libwebpdecoder3 libwmf-0.2-7
Reading package lists... Done
Building dependency tree... Done
Suggested packages:
  libwmf-0.2-7-gtk
  libgirepository-1.0-1 libidn12 libmemcachedutil2t64 libwebpdecoder3
  libwmf-0.2-7
0 upgraded, 5 newly installed, 0 to remove and 2 not upgraded.
Need to get 328 kB of archives.
After this operation, 1278 kB of additional disk space will be used.
…
```

heroku-24/setup.sh
```diff
@@ -119,6 +123,12 @@ packages=(
   zstd
 )
 
+if [[ $(dpkg --print-architecture) == "amd64" ]]; then
+  packages+=(
+    libgirepository-1.0-1 # Because libgirepository1.0-dev and libgirepository-1.0-dev are in the build image
+  )
+fi
```
edmorley, line 130: Prior to this architecture-conditional addition, the ARM64 and AMD64 **run** images contained exactly the same packages (with the exception of the binutils package name including the arch so being named `binutils-x86-64-linux-gnu` vs `binutils-aarch64-linux-gnu` - but effectively the package provides the same functionality). Now their contents will vary by arch. 

Whilst I know that (a) the run images containing the same packages isn't something that we can necessarily guarantee for the future (eg Heroku-26), and (b) the **build** images already vary by arch (in part due to `libgirepository-1.0-dev`), but there is still some appeal/simplicity to me of the packages list being the same - if only to make things simpler when we display the table of packages on Dev Center (and to avoid user confusion when they try to work out what's available for their app, without them having to know what arch they are using).

Given that `libgirepository-1.0-1` is only needed because of `libgirepository-1.0-dev`, and the headers are only pulled in because of slightly obscure reasons (it's a transitive dependency of `libmagickwand-dev` and friends on one arch only), I wonder if we can just skip installing this package? I don't think it's the type of package anything will depend on at runtime. (Plus it's not going to be available on ARM64 anyway)

heroku-22/setup.sh
```diff
@@ -73,6 +73,7 @@ packages=(
   libgd3
   libgdk-pixbuf2.0-0
   libgdk-pixbuf2.0-common
+  libgeoip1
```
edmorley, line 76: For Heroku-24 I removed `libgeoip-dev`, `libgeoip1`, `geoip-database` etc for the reason that they were (a) broken on older stacks, (b) clearly not used by anyone, since no one had reported the breakage.

Given they aren't supported on Heroku-24, should we skip adding them back to older stacks? If anyone actually does start depending on those packages, they are only going to then have to find another solution (eg Apt buildpack) regardless when they upgrade to Heroku-24 otherwise.

(This package is very small, so this isn't about the size, more us being clear about intent and easing migration to newer stacks etc)

edmorley (commented):
Thank you for listing the size impacts - makes review easier!

I'm happy for us to add these - just have a few questions/comments first :-)

heroku-24/setup.sh
dzuelke: Yeah I prioritized consistency between run and build image over consistency across architectures here, since the build images already differ, so we have the problem for the Dev Center packages list already.

I dug through the various Debian control files for a bit to see if the `libgirepository-1.0-dev` dependency difference between architectures is even intentional, but couldn't find something conclusive in a reasonable amount of time.

Happy to drop this one instead.

heroku-22/setup.sh
dzuelke: Sounds reasonable.

edmorley (approved):
Thank you!

edmorley:
@dzuelke Could you update the PR description so it matches the final version of the PR? :-)

PR #320, opened 2024-08-09
Publish `.nightly` tags nightly

joshwlewis:
Currently, the `.nightly` tags (e.g. `heroku/heroku:24.nightly`) are only published when we merge things to main. When no changes happen on `main` for a while, the `.nightly` tags might actually be behind the non-nightly tags. 

This updates the CI workflow to run Monday-Friday at midnight.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ on:
     schedule:
       - cron: "0 0 * * 1-5"
   pull_request:
+  workflow_dispatch:
 
 permissions:
   contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,8 @@ on:
       - main
     tags:
       - v*
+    schedule:
+      - cron: "0 0 * * 1-5"
   pull_request:
 
 permissions:


.github/workflows/ci.yml
```diff
@@ -7,6 +7,8 @@ on:
       - main
     tags:
       - v*
+    schedule:
+      - cron: "0 0 * * 1-5"
   pull_request:
```
edmorley, line 13: (optional) Should we also have a `workflow_dispatch` here, so we can manually trigger if needed? eg similar to:
https://github.com/heroku/cnb-builder-images/blob/0a8816f18165c7f14e8841b3b931bb67a9d351a5/.github/workflows/build-test-publish.yml#L9

```suggestion
  workflow_dispatch:

```

edmorley (approved):
Thank you!

.github/workflows/ci.yml
```diff
@@ -7,7 +7,10 @@ on:
       - main
     tags:
       - v*
+    schedule:
+      - cron: "0 0 * * 1-5"
```
edmorley, line 11: @joshwlewis I was wondering why the nightly images weren't being updated - turns out this YAML block is over-indented and so being ignored. (If I open this workflow file in VSCode, where I have the GitHub Actions extension enabled, I get a schema validation lint warning.)

PR #321, opened 2024-08-14
Add type-urw-base35 package to heroku-24

dzuelke:
For example, the following command does not work anymore:

    convert -background white -fill purple -font Courier -pointsize 72 label:Heroku test.png

This is a pain to work around using the Apt buildpack, because the base `/etc/ImageMagick-6/type.xml` includes more specific mappings for the `gsfonts` and `type-urw-base35` package's font files with absolute paths, so users would have to produce a huge custom `type.xml` with changed paths for each font in the package, and then point ImageMagick to that config using the `MAGICK_FONT_PATH` environment variable.

GUS-W-16494478

### DIFF
--- a/heroku-24/setup.sh
+++ b/heroku-24/setup.sh
@@ -43,6 +43,8 @@ packages=(
   bzip2
   curl
   file
+  fonts-dejavu-core
+  fonts-dejavu-mono
   fonts-urw-base35 # ImageMagick's default type.xml config points to these, includes e.g. Courier or Helvetica
   gettext-base # For `envsubst`.
   gir1.2-harfbuzz-0.0 # Used by FFmpeg in heroku-buildpack-activestorage-preview.
--- a/heroku-24/setup.sh
+++ b/heroku-24/setup.sh
@@ -43,6 +43,7 @@ packages=(
   bzip2
   curl
   file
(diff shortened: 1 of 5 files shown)


heroku-24/setup.sh
```diff
@@ -43,6 +43,7 @@ packages=(
   bzip2
   curl
   file
+  fonts-urw-base35 # ImageMagick's default type.xml config points to these, includes e.g. Courier or Helvetica
```
joshwlewis, line 46:
```suggestion
  fonts-dejavu-core
  fonts-dejavu-mono
  fonts-urw-base35 # ImageMagick's default type.xml config points to these, includes e.g. Courier or Helvetica
```

joshwlewis (approved):

heroku-24/setup.sh
```diff
@@ -43,6 +43,9 @@ packages=(
   bzip2
   curl
   file
+  fonts-dejavu-core
+  fonts-dejavu-mono
```
edmorley, line 47: Please could you add comments to these? In general we should have comments for all non-obvious packages in this list.

Also, the `fonts-dejavu-mono` package is a transitive dependency of `fonts-dejavu-core`, and we now intentionally avoid listing transitive dependencies here - so could you remove `fonts-dejavu-mono` from this list?

PR #322, opened 2024-08-19
Add CTC check prior to tag promotion

joshwlewis:
This adds a CTC check prior to promoting to main tags e.g. from `heroku/heroku:24.v133` to `heroku/heroku:24`. This should prevent releases during deployment moratoriums.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,7 +204,7 @@ jobs:
     steps:
       - name: Obtain CTC Lock via TPS API
         run: |
-          curl -vv -X PUT \
+          curl -sS -X PUT \
             -H "ACCEPT: application/json" \
             -H "Content-Type: application/json" \
             -H "Authorization: Token ${{ secrets.TPS_TOKEN }}" \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,9 +207,9 @@ jobs:
           curl -vv -X PUT \
             -H "ACCEPT: application/json" \
             -H "Content-Type: application/json" \
-            -H "Authorization: Token ${{ secrets.tps_token }}" \
+            -H "Authorization: Token ${{ secrets.TPS_TOKEN }}" \
             -d '{"lock": {"sha": "${{ github.sha }}", "component_name": "${{ vars.tps_component }}"}}' \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -197,10 +197,25 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Obtain CTC Lock via TPS API
+        run: |
+          curl -sS -X PUT \
+            -H "ACCEPT: application/json" \
+            -H "Content-Type: application/json" \
+            -H "Authorization: Token ${{ secrets.TPS_TOKEN }}" \
+            -d '{"lock": {"sha": "${{ github.sha }}", "component_name": "${{ vars.tps_component }}"}}' \
+            ${{ secrets.TPS_CTC_API_URL }}
```
dzuelke, line 212:
```suggestion
            "${{ secrets.TPS_CTC_API_URL }}"
```

dzuelke (approved):

.github/workflows/ci.yml
```diff
@@ -197,10 +197,25 @@ jobs:
...
           done
 
+  ctc-check:
+    name: Obtain CTC Lock
+    if: github.ref_type == 'tag'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Obtain CTC Lock via TPS API
+        run: |
+          curl -sS -X PUT \
```
edmorley, line 207: @joshwlewis This check currently won't actually block releases, since CTC uses non-HTTP-200 responses to denote blocking a lock (https://github.com/heroku/tps/blob/master/docs/ctc.md#responses), and Curl ignores these unless the `--fail` argument is used.

eg compare to the Cheverny implementation, which checks for HTTP 200/201:
https://github.com/heroku/cheverny/blob/f27b132b8b829b16d2161977f00316faf0f06520/lib/mediators/stack/ctc_checker.rb#L27

For bonus points, we should probably also use `--fail-with-body` rather than `--fail`, since the former will also print the API response message:
https://curl.se/docs/manpage.html#--fail-with-body

edmorley: For even more bonus points, perhaps we should add `--connect-timeout` and `--retry` to reduce flakiness?

PR #323, opened 2024-08-27
Fail CTC check on 400+ responses

joshwlewis:
As @edmorley pointed out in https://github.com/heroku/base-images/pull/322#discussion_r1732576122, the CTC check won't actually fail as is. I've added `--fail-with-body` to curl so that the command does fail with a non-zero exit if the HTTP status code is 400 or higher.

I've also added some retries to reduce flakiness.

GUS-W-16376259.


edmorley (approved):
Thank you :-)

PR #327, opened 2024-09-04
Update manifest.yml to use tag rather than sha

joshwlewis:
The current iteration of https://s3.amazonaws.com/heroku_stacks_production/manifest.yml.signed currently includes entries with git shas like this:

```
git_ref: b4689bb306581b71b29701c55ecd9c6efb730d9c
```

Prior to https://github.com/heroku/base-images/pull/318, git tags were used instead. e.g.:

```
git_ref: v105
```

This PR updates the CI action to use git tags, as before.

This PR additionally updates the CTC to use git tags too, since it's common to successive base images against the same sha.

GUS-W-17066497.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
           path: /tmp/heroku-${{ matrix.stack-version }}*
       - name: Upload heroku runtime images to staging
         run: |
-          bin/upload-runtime-images.sh ${{ matrix.stack-version }} ${{ github.sha }}
+          bin/upload-runtime-images.sh ${{ matrix.stack-version }} ${{ github.ref_name }}
 
   publish-images:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -99,7 +99,7 @@ jobs:
           path: /tmp/heroku-${{ matrix.stack-version }}*
       - name: Upload heroku runtime images to staging
         run: |
-          bin/upload-runtime-images.sh ${{ matrix.stack-version }} ${{ github.sha }}
+          bin/upload-runtime-images.sh ${{ matrix.stack-version }} ${{ github.ref_name }}
```
edmorley, line 102: The `ctc-check` job also uses `github.sha` (when iirc that used the tag before too). Should that be switched back to the tag too?

edmorley: @joshwlewis Just checking you saw ^ ? :-)

PR #329, opened 2024-10-29
feature: add the AWS RDS cert bundles

coreypurcell:
Reopening https://github.com/heroku/base-images/pull/328

This PR will add RDS CA bundle to the stacks. This will allow customers to set their postgres connections to verify-full when connecting to a essential tier database. As requested the bundle is now included in the repo instead of downloading. You can follow my pattern below to retest, if desired, using an essential-0 postgres.

Testing
Before changes

root@aecdd0b8a9ae / [heroku/heroku:24]
docker > psql 'postgres://udkn8531065hs2:<redacted>@c4fuk4g4mva0ni.cluster-cmgvywijhzgr.us-east-1.rds.amazonaws.com:5432/d7434taaso5gbu?sslmode=verify-full&sslrootcert=system'
psql: error: connection to server at "c4fuk4g4mva0ni.cluster-cmgvywijhzgr.us-east-1.rds.amazonaws.com" (100.64.1.39), port 5432 failed: SSL error: certificate verify failed
Built the images
Ran this pattern in all 3 images

docker run -it --rm --platform=linux/amd64 33f6435378c7 bash
root@13d8683a53ca:/# psql 'postgres://udkn8531065hs2:<redacted>@c4fuk4g4mva0ni.cluster-cmgvywijhzgr.us-east-1.rds.amazonaws.com:5432/d7434taaso5gbu?sslmode=verify-full&sslrootcert=system'
psql (16.4 (Ubuntu 16.4-1.pgdg22.04+2), server 15.7)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
Type "help" for help.

root@6cf200a37298:/# curl -s https://cdn.azul.com/zulu/bin/zulu17.40.19-ca-jdk17.0.6-linux_x64.tar.gz | tar xz
root@6cf200a37298:/# zulu17.40.19-ca-jdk17.0.6-linux_x64/bin/keytool -list -v -keystore /etc/ssl/certs/java/cacerts | grep "Amazon RDS" | grep "Owner" | wc -l
Enter keystore password:
...

100


heroku-20/rds-global-bundle.pem
```diff
@@ -0,0 +1,2528 @@
+-----BEGIN CERTIFICATE-----
```
joshwlewis, line 1: This looks like the same file copied 3 times? Is there any way to put it into the repo only once? Maybe a `certs` or `share` directory at the root?

edmorley: I believe the Dockerfile context root is the stack subdirectory, so using a shared asset outside of that wouldn't be possible without some restructuring (which I'm not against) - which is why I added the Postgresql APT key assets multiple times in #256. Git does de-dupe identical files within its store, so it won't bloat the repo at least.

heroku-20/setup.sh
```diff
@@ -160,6 +160,16 @@ apt-get install -y --no-install-recommends "${packages[@]}"
 
 cp /build/imagemagick-policy.xml /etc/ImageMagick-6/policy.xml
 
+# Install AWS RDS global CA bundle (https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.CertificatesAllRegions)
+mkdir -p /usr/local/share/ca-certificates/rds-ca-certs
+pushd /usr/local/share/ca-certificates/rds-ca-certs
```
edmorley, line 165: I think the pushd/popd might be redundant given the full path is used inside the awk script? :-)

joshwlewis (approved):

PR #332, opened 2024-11-12
Fixing reproducibility error introduced by 24-CNB vs 24-Heroku

usiegj00:
The 24-Heroku stack adds a user like `u12345` and a `dyno` user, both with `HOME` of `/app`. The 24-CNB stack breaks this convention by introducing a `heroku` user with a `HOME` of `/home/heroku`. This prevents packages such as [fontconfig which expects .fonts to be available in HOME](https://www.freedesktop.org/software/fontconfig/fontconfig-user.html). 

In non-CNB builds, `HOME` and `/app` are the same, so a `/app/.fonts/this.ttf` is available as expected at runtime.

In CNB builds, these are different, so the same font would be at `/app/.fonts/this.ttf`, while the runtime path searched would be `/home/heroku/.fonts`.

This fix has been thoroughly tested and helps Heroku customers achieve better reproducibility and alignment with open standards.

### DIFF
--- a/heroku-24-build/Dockerfile
+++ b/heroku-24-build/Dockerfile
@@ -14,3 +14,7 @@ ENV CNB_GROUP_ID=1000
 # Stack IDs are deprecated, but we still set this for backwards compatibility:
 # https://github.com/buildpacks/spec/blob/platform/0.13/platform.md#iobuildpacksstack-labels
 ENV CNB_STACK_ID="heroku-24"
+
+# Match HOME operation done in [heroku-22](https://github.com/usiegj00/base-images/blob/3bed84a6cf88bc6535b1ef18acbc454f88c854d8/heroku-22-cnb/Dockerfile#L17). 
+# This is related to [the user HOME issue exposed here](https://github.com/heroku/base-images/pull/332).
+ENV HOME=/workspace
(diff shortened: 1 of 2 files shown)


edmorley (commented):
Hi! 

Sorry for the delay in replying - we were focusing on work leading up to the recent Heroku Fir announcement:
https://blog.heroku.com/tags/nextgen

Thank you for taking a look at these images and opening the PR. 

Unfortunately we won't be able to merge these changes since the differences you mention between the images are by design - the Heroku-24 behaviour is the preferred behaviour. 

If anything we'd want to change the old images to match rather than vice versa - however, attempting to unify the behaviour across older and newer stacks (in either direction) would be a breaking change - and we cannot typically make breaking changes to stacks once released.

I would recommend reading the CNB spec for understanding the direction the standards are heading:
https://github.com/buildpacks/spec

There's also prior discussion on the `HOME` location trade-offs in:
https://github.com/buildpacks/spec/issues/186
https://github.com/heroku/cnb-shim/pull/23

In general, it's expected that there will be some differences between Cedar generation apps (using classic buildpacks) and Fir generation apps (using CNBs), since we are aligning on best practices and open standards. If there are specific cases where we can improve parity by backporting a change to classic in a non-breaking way, then please do let us know :-)

edmorley:
Closing since this change isn't one we want to make (for the reasons above).

PR #340, opened 2025-01-28
Add tzdata-legacy to Heroku-24

joshwlewis:
Ubuntu 24.04's `tzdata` package does not include older/non-standard timezones like `US/Eastern` and `US/Central`, where Ubuntu 22.04 does. It's mentioned in the release notes [here](https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890#p-99950-tzdata-package-split). Therefore, this is also a change between Heroku-22 and Heroku-24.

Some (mostly Rails) users have these legacy timezone names stored in their application's database, which makes upgrading to Heroku-24 problematic. `tzdata-legacy` adds the removed timezones.

Note that the `tzdata-legacy` package can't be correctly installed via a buildpack (classic or CNB), since tools like `zic` and `zdump` expect the data to be in `/usr/share/zoneinfo`, which isn't writeable during builds or runtime.

I'm not sure if this is totally necessary. Ruby users can alternatively get the legacy timezones via the `tzinfo-data` RubyGem. The `tzinfo-data` gem doesn't solve the problem outside of Ruby, but I haven't seen this problem reported outside of a Ruby context.


schneems:
Cross linking (internal only access) tickets:

- https://heroku.support/1516496
- https://heroku.support/1458204
- https://heroku.support/1513626 

I'm good with adding this in as it stops the initial pain. 

Here's an example error message (to possibly help with people finding the fix to use the `tzdata-legacy` gem):

```
~ $ ruby -r"active_support/all" -e "puts Time.now.in_time_zone 'US/Central'"
/app/vendor/bundle/ruby/3.4.0/gems/activesupport-8.0.1/lib/active_support/core_ext/time/zones.rb:84:in 'Time.find_zone!': Invalid Timezone: US/Central (ArgumentError)

      ActiveSupport::TimeZone[time_zone] || raise(ArgumentError, "Invalid Timezone: #{time_zone}")
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	from /app/vendor/bundle/ruby/3.4.0/gems/activesupport-8.0.1/lib/active_support/core_ext/date_and_time/zones.rb:21:in 'DateAndTime::Zones#in_time_zone'
	from -e:1:in '<main>'
```

I think the only possible concern I have is considering how we get off this train once we're on it. i.e. what the path to detection, deprecation, and removal might look like in the future. That sounds like a relatively small problem for future-us so I'm in favor of moving forward with adding it here. Saving approval for Ed to have him look over it in the morning.

schneems:
## Go

Fails in the same way

## Python

From https://docs.python.org/3/library/zoneinfo.html

>  The zoneinfo module does not directly provide time zone data, and instead pulls time zone information from the system time zone database or the first-party PyPI package tzdata, if available. Some systems, including notably Windows systems, do not have an IANA database available, and so for projects targeting cross-platform compatibility that require time zone data, it is recommended to declare a dependency on tzdata. If neither system data nor tzdata are available, all calls to ZoneInfo will raise ZoneInfoNotFoundError.

Timezones don't work out of the box and the docs say to use the library (which I've not tried, but I am guessing vendors the data rather than relying on the system)

## Node.js

Does the conversion for you:

```
> process.env.TZ = "US/Central"
'US/Central'
> Intl.DateTimeFormat().resolvedOptions().timeZone
'America/Chicago'
```

edmorley (commented):
Switching this to draft for now to make its status clearer and stop it appearing in our review queue :-)

Since the docs were added to the Heroku-24 migrating guide I haven't seen any further support tickets about this fwiw. (Though I'm still fine with adding if we think it's best.)

edmorley:
We've had no further requests for this, so closing for now.

PR #345, opened 2025-02-22
Update installed packages

runesoerensen:
This reverts most of the "installed package list" changes introduced in https://github.com/heroku/base-images/pull/333, with the exception of `heroku-20-build` [which installs the `postgresql-server-dev-16` package](https://github.com/heroku/base-images/blob/090a999cda061ad91cf59120246df6eff1fe5f54/heroku-20-build/setup.sh#L80) - and instead adds the `postgresql-common-dev` package, which depend on `libio-pty-perl` and `libipc-run-perl`.

See https://www.postgresql.org/message-id/E1tiZyA-008IKN-8q%40atalia.postgresql.org for more details.

### DIFF
--- a/heroku-20-build/installed-packages-amd64.txt
+++ b/heroku-20-build/installed-packages-amd64.txt
@@ -568,6 +568,7 @@ poppler-utils
 postgresql-client-16
 postgresql-client-common
 postgresql-common
+postgresql-common-dev
 postgresql-server-dev-16
 procps
 python-is-python3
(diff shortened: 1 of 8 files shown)


edmorley (approved):
Given these packages were only recently pulled in via transitive deps (Nov 2024; #333) and are for obscure Perl library modules customer apps are unlikely to depend upon, I'm fine with updating the package list here to match the new reality (after the latest Postgres package upstream shuffle) rather than explicitly adding the packages back.

PR #346, opened 2025-02-22
Add `liblttng-ust1` to Heroku-22 and Heroku-24

runesoerensen:
This PR adds the `liblttng-ust1` package (listed as an Ubuntu [22.04](https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-install?tabs=dotnet9&pivots=os-linux-ubuntu-2204#dependencies-1) and [24.04](https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-install?tabs=dotnet9&pivots=os-linux-ubuntu-2404#dependencies-1) .NET runtime dependency) to Heroku-22 and Heroku-24.

### DIFF
--- a/heroku-22-build/installed-packages-amd64.txt
+++ b/heroku-22-build/installed-packages-amd64.txt
@@ -305,6 +305,9 @@ liblqr-1-0-dev
 liblsan0
 libltdl-dev
 libltdl7
+liblttng-ust-common1
+liblttng-ust-ctl5
+liblttng-ust1
 liblz4-1
 liblz4-dev
 liblzf-dev
--- a/heroku-22/installed-packages-amd64.txt
+++ b/heroku-22/installed-packages-amd64.txt
@@ -204,6 +204,9 @@ liblmdb0
 liblqr-1-0
 liblsan0
 libltdl7
+liblttng-ust-common1
+liblttng-ust-ctl5
+liblttng-ust1
 liblz4-1
 liblzf1
 liblzma5
--- a/heroku-22/setup.sh
+++ b/heroku-22/setup.sh
@@ -85,6 +85,7 @@ packages=(
   libharfbuzz0b
   libhashkit2
   libheif1
+  liblttng-ust1
   liblzf1
   libmagickcore-6.q16-3-extra
   libmcrypt4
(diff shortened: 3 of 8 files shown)


edmorley (approved):
Given these packages are small (1-2 MB total) and needed for a primary language, I agree we should add these :-)

I'm curious though - have these always been requirements for .NET? If so, how did you discover they were missing now / what was the impact? Is this something we should have test .NET coverage for? 

eg for Python I use an `ldd` check both at binary compile time (which doesn't apply to .NET) and as part of my integration tests:

- https://github.com/heroku/heroku-buildpack-python/blob/53a59ae3cb7f0bd7b55d5bdcc2005989b2d3825b/builds/test_python_runtime.sh#L44-L50
- https://github.com/heroku/buildpacks-python/blob/9754ef3154f1907a9686b59b8a777b23f71283d9/tests/python_version_test.rs#L128-L133

edmorley:
If some of the libraries aren't showing up in the output of `ldd`, then it may mean they are loaded using `dlopen` instead of being dynamically linked at compile time (xref https://stackoverflow.com/questions/45388186/what-are-the-use-cases-of-dlopen-vs-standard-dynamic-linking).

runesoerensen:
@edmorley I've removed the commit that added `libunwind` for the time being. While it's listed as an Ubuntu dependencies, it doesn't seem to be required for any of the architectures we support (it was vendored in years ago).

runesoerensen:
> I like that approach a lot (and stole it for [the draft PR](https://github.com/heroku/buildpacks-dotnet/pull/205/commits/beca3e381b1c36fc7fb01f434f6d191bb7328a09) :-)). It shows liblttng-ust.so.0 missing on both Heroku-22 + Heroku-24.

> If some of the libraries aren't showing up in the output of `ldd`, then it may mean they are loaded using `dlopen` instead of being dynamically linked at compile time (xref https://stackoverflow.com/questions/45388186/what-are-the-use-cases-of-dlopen-vs-standard-dynamic-linking).

Looks like we might not be able to verify `liblttng` using the dynamically linked libraries check implemented in both buildpacks (due to https://github.com/dotnet/runtime/issues/57784). We have `liblttng-ust.so.1` with the change introduced by this PR, but `libcoreclrtraceptprovider.so` looks for `liblttng-ust.so.0`.

PR #348, opened 2025-03-17
Only expose the manifest token to the workflow step that needs it

edmorley:
As part of security-hardening our GHA workflows, this moves the manifest token/url env vars to the specific workflow step that uses it, rather than exposing it to all steps within that job.

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions

GUS-W-18053749.


runesoerensen (approved):

runesoerensen:
@edmorley looks good -- just needs to be rebased :)

PR #357, opened 2025-08-01
Add `libtirpc3t64` to Heroku-24 packages

colincasey:
Changes to the installed packages list for Heroku-24 (amd64 and arm64) are causing CI failures: https://github.com/heroku/base-images/actions/runs/16662909221/job/47163740980

GUS-W-19212320

### DIFF
--- a/heroku-24-build/setup.sh
+++ b/heroku-24-build/setup.sh
@@ -53,7 +53,6 @@ packages=(
   patchelf
   python3 # Often needed during the building of non-Python apps. e.g. For Node.js packages that use node-gyp.
   zlib1g-dev
-  libtirpc3t64 # used to be a dependency of iproute2 (e.g. /bin/ip linked against libtirpc.so.3), now requiring explicitly to prevent package removal - can be removed for heroku-26
 )
 
 apt-get update --error-on=any
--- a/heroku-24-build/setup.sh
+++ b/heroku-24-build/setup.sh
@@ -53,6 +53,7 @@ packages=(
   patchelf
   python3 # Often needed during the building of non-Python apps. e.g. For Node.js packages that use node-gyp.
   zlib1g-dev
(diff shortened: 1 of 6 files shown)


dzuelke (changes requested):
Unlikely to cause real-life issues for people, but technically, this qualifies as a regression. Our policy is to never remove packages, since existing code may have linked to it.

If a transient dependency like this gets dropped, we need to instead explicitly add it to the list of packages. I'd suggest adding it to the end of the `packages` list in `heroku-24/setup.sh` with an explicit comment, like so:

```bash
packages=(
  …
  libtirpc3t64 # used to be a dependency of iproute2 (e.g. /bin/ip linked against libtirpc.so.3), now requiring explicitly to prevent package removal - can be removed for heroku-26
)
```

dzuelke (changes requested):
You added it to `heroku-24-build/setup.sh`, but it needs to be in `heroku-24/setup.sh` :)

heroku-24-build/setup.sh
```diff
@@ -53,6 +53,7 @@ packages=(
   patchelf
   python3 # Often needed during the building of non-Python apps. e.g. For Node.js packages that use node-gyp.
   zlib1g-dev
+  libtirpc3t64 # used to be a dependency of iproute2 (e.g. /bin/ip linked against libtirpc.so.3), now requiring explicitly to prevent package removal - can be removed for heroku-26
```
dzuelke, line 56:
```suggestion
  libtirpc3t64 # used to be a dependency of iproute2 (e.g. /bin/ss linked against libtirpc.so.3), now requiring explicitly to prevent package removal - can be removed for heroku-26
```

Had the wrong binary there, sorry ;)

colincasey: Fixed in https://github.com/heroku/base-images/pull/357/commits/24806357ef78332b24fa731de07876b51538f964

dzuelke (approved):

PR #361, opened 2025-09-01
Add `media-types` package to Heroku-24 run-image

### LINKED ISSUE #360: Images based on Ubuntu 24.04 are missing `/etc/mime.types` (`media-types` package)
(Apologies if this isn’t the right place for this issue – I’m not very familiar with the different components of Heroku.)

Images based on Ubuntu 24.04, compared to ones based on Ubuntu 22.04, are missing the `/etc/mime.types` file (shipped by the `media-types` APT package). This file contains MIME types for various file extensions, and is used by different servers to send an appropriate `Content-Type` response header. Without this file, some servers can’t send a `Content-Type`, which may break an application; for example, browsers will refuse to load `<script type="module">` scripts if the response is missing a `text/javascript` content type:

> Loading module from “https://lucaswerkmeister-test.toolforge.org/index.js” was blocked because of a disallowed MIME type (“”). [Firefox]
> Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec. [Chromium]

I have experienced this issue with Python-based servers (specifically uWSGI and Gunicorn), which read the file via the [mimetypes](https://github.com/python/cpython/blob/v3.13.7/Lib/mimetypes.py#L48-L58) module, but I expect other servers and languages may also be affected.

Steps to reproduce:

- Clone the `heroku-mime-types` branch of [this repository](https://gitlab.wikimedia.org/toolforge-repos/lucaswerkmeister-test/): `git clone -b heroku-mime-types https://gitlab.wikimedia.org/toolforge-repos/lucaswerkmeister-test.git && cd lucaswerkmeister-test`
- Create an image based on Ubuntu 22.04: `pack-cli build --builder docker.io/heroku/builder:22 myimage`
- Run the image: `docker run --rm -e PORT=8000 -p 8000:8000 myimage`
- Check the response headers for a JS file in the app: `curl -I http://localhost:8000/index.js` (you’ll need to do this in another terminal as the original one will have Docker running in the foreground) – you should see `Content-Type: text/javascript`.
- Stop the `docker run` with <kbd>Ctrl</kbd>+<kbd>C</kbd>.
- Create an image based on Ubuntu 24.04: `pack-cli build --builder docker.io/heroku/builder:24 myimage`
- Run the image: `docker run --rm -e PORT=8000 -p 8000:8000 myimage` (you should see a message “!!! no /etc/mime.types file found !!!” in the uWSGI initialization output)
- Check the response headers for a JS file in the app: `curl -I http://localhost:8000/index.js` (another terminal, again) – this time the response will contain no `Content-Type` header.

Depending on the server, you may be able to work around this by specifying a custom `mime.types` file (e.g. [uWSGI `--mime-file` option](https://uwsgi.readthedocs.io/en/latest/StaticFiles.html#mime-types), [example workaround](https://gitlab.wikimedia.org/toolforge-repos/wdactle/-/commit/69ac5a1656)) or sending the right header in your application code. But ideally the `/etc/mime.types` file should be part of the base image.

(This issue is also tracked on the Wikimedia Toolforge platform as [T403303](https://phabricator.wikimedia.org/T403303).)

edmorley:
Since it provides `/etc/mime.types`, which is one of the locations used by the Python `mimetypes` stdlib module:
https://github.com/python/cpython/blob/v3.13.7/Lib/mimetypes.py#L48-L58

In Heroku-22 and older the package is included due to it being a transitive dependency of other packages (such as system Python), however, for Heroku-24's smaller packages list we now need to include it explicitly.

The package is ~90KB.

Fixes #360.
GUS-W-19469378.

### DIFF
--- a/heroku-24/installed-packages-amd64.txt
+++ b/heroku-24/installed-packages-amd64.txt
@@ -312,6 +312,7 @@ login
 logsave
 lsb-release
 mawk
+media-types
 mlock
 mount
 mysql-common
--- a/heroku-24/installed-packages-arm64.txt
+++ b/heroku-24/installed-packages-arm64.txt
@@ -312,6 +312,7 @@ login
 logsave
 lsb-release
 mawk
+media-types
 mlock
 mount
 mysql-common
--- a/heroku-24/setup.sh
+++ b/heroku-24/setup.sh
@@ -106,6 +106,7 @@ packages=(
   libzip4 # Used by the PHP runtime.
   locales
   lsb-release
+  media-types # Provides /etc/mime.types, used by Python's `mimetypes` stdlib module.
   nano # More usable than ed but still much smaller than vim.
   netcat-openbsd
   openssh-client # Used by Heroku Exec.


Malax (approved):

edmorley:
This will be available in the next image release, which will be sometime next week:
https://github.com/heroku/base-images/releases
https://devcenter.heroku.com/articles/stack-update-policy#updates-to-existing-stacks

PR #387, opened 2026-05-12
Update postgresql-client to v18

troycoll:
Updates the Postgres client packages in all base images versions from v17 to v18.

See:
- https://www.postgresql.org/about/news/postgresql-18-released-3142/
- https://www.postgresql.org/docs/release/18.0/
- https://www.postgresql.org/docs/release/18.1/
- https://www.postgresql.org/docs/release/18.2/
- https://www.postgresql.org/docs/release/18.3/

GUS-W-21857354.

### DIFF
--- a/heroku-22-build/installed-packages-amd64.txt
+++ b/heroku-22-build/installed-packages-amd64.txt
@@ -565,7 +565,7 @@ pinentry-curses
 pkg-config
 poppler-data
 poppler-utils
-postgresql-client-17
+postgresql-client-18
 postgresql-client-common
 procps
 python-is-python3
--- a/heroku-22/installed-packages-amd64.txt
+++ b/heroku-22/installed-packages-amd64.txt
@@ -371,7 +371,7 @@ perl-modules-5.34
 pinentry-curses
 poppler-data
 poppler-utils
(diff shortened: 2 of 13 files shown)


edmorley (commented):
@troycoll Thank you for the PR! 

Thanks for tracking down the package churn -- in general we try to avoid removing packages from a base image once GA to avoid breaking apps, so in cases like these where a transitive dependency would otherwise be removed, we add it back as an explicit dependency instead.

I've opened #388 for that - could you rebase on that? :-)

edmorley (approved):
Thank you for updating this - checking the PG18 release notes now I don't see any breaking changes that should affect us, so we should be good to update :-)

edmorley:
@troycoll We typically release base image updates every two weeks. The most recent release was last week, so the next regular release will be next week - however, we could expedite one if needed. Once a release is started it takes a few days to roll out to the Common Runtime/Fir, then at least a few weeks to roll out to Dogwood (though runtime may be able to expedite this if needed).

edmorley:
Base image release started in https://github.com/heroku/base-images/releases/tag/v176 / GUS-W-20977780

PR #391, opened 2026-05-21
Add stack-package-report tool

Malax:
Part of the Cheverny replacement. Adds a new Rust tool, `stack-package-report`, that generates the same changelog markdown and DevCenter package tables that Cheverny produces today.

The TSV package-list inputs the tool consumes will be produced in CI via a follow-up PR; this one is split out to keep the review small.

The tool works locally and is thoroughly tested — roughly half the code is unit tests.

Also adds `.github/workflows/rust-ci.yml` (clippy, rustfmt, unit tests) and a matching `cargo` entry in `.github/dependabot.yml`.


stack-package-report/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+use indoc::indoc;
+
+use crate::package_changelog_markdown::render_package_changelog_markdown;
+use crate::package_list_devcenter::render_package_list_devcenter;
+use crate::stack_package_list::parse_stack_package_list_dir;
+
+fn main() {
+    let args: Vec<String> = std::env::args().collect();
+
+    let exit_code = match args.get(1).map(String::as_str) {
```
colincasey, line 20: minor - why not use clap? i get that the arg parsing here is light so why add the extra dependency but i find, even with small CLI tools, using clap pays off vs. hand-rolling it + we use it in plenty of other places.

stack-package-report/src/stack_package_list.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn parse_stack_package_list_dir_yields_one_entry_per_tsv_file() {
+        let dir = make_temp_dir();
+        fs::write(
+            dir.join("heroku-24_linux-amd64-packages.tsv"),
+            "bash\t5.1\nbinutils-x86-64-linux-gnu\t2.38\n",
+        )
+        .unwrap();
+        fs::write(
+            dir.join("heroku-24-build_linux-arm64-packages.tsv"),
+            "bash\t5.1\nbinutils-aarch64-linux-gnu\t2.38\n",
```
colincasey, line 145: is it possible for different architectures to contain different versions for a package? e.g.;
- binutils-x86-64-linux-gnu@2.38
- binutils-aarch64-linux-gnu@2.38.1

stack-package-report/src/stack_package_list_id.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+#[derive(Copy, Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub(crate) enum Architecture {
+    Amd64,
+    Arm64,
+}
+
+impl StackPackageListId {
+    #[allow(unused)]
```
colincasey, line 31: is this used anywhere besides tests?

stack-package-report/src/package_list_devcenter.rs
```diff
@@ -0,0 +1,523 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::stack_package_list::StackPackageList;
+    use crate::stack_package_list_id::{Architecture, Stack, StackPackageListId};
+    use indoc::indoc;
+
+    #[test]
+    #[allow(clippy::too_many_lines)]
+    fn renders_full_devcenter_table_covers_all_features() {
```
colincasey, line 108: do you see any value in extending this test to verify it produces the exact content that's currently on devcenter? that would be a decent proof of parity.

.github/workflows/rust-ci.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  unit-test:
+    name: Unit Tests
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        working-directory: stack-package-report
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
colincasey:
```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

stack-package-report/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[dependencies]
+indoc = "2.0.7"
+regex = "1"
+thiserror = "2.0.18"
+
+[lints.clippy]
+all = { level = "warn", priority = -1 }
+pedantic = { level = "warn", priority = -1 }
+format_push_string = "allow"
```
colincasey, line 14: do we disable this elsewhere? why do it here when there's a recommended alternative? 
https://rust-lang.github.io/rust-clippy/stable/index.html#format_push_string

stack-package-report/src/package_list_devcenter.rs
```diff
@@ -0,0 +1,523 @@
...
+                    ("sudo", "456"),
+                    ("sudo-dev", "456"),
+                    ("syslinux", "3:6.04~git20190206.bf6db5b4+dfsg1-3ubuntu1"),
+                ],
+            ),
+        ];
+
+        let stack_order = vec!["heroku-22".to_string(), "heroku-24".to_string()];
+
+        let expected = indoc! {r#"
```
colincasey, line 222: this type of thing is perfect for `insta`. the test would be a lot easier to read and, if the output changes, review for what's wrong.

stack-package-report/src/package_list_devcenter.rs
```diff
@@ -0,0 +1,523 @@
...
+    groups
+}
+
+// Some Ubuntu package versions contain a long `git…` segment whose only
+// internal separators are `+` or `.` (e.g. `0.5.11+git20210903+057cd650a4ed-3build1`).
+// Browsers treat that as one unbreakable token and the table column blows
+// out horizontally. Inserting a zero-width space after the `git…` segment
+// gives the browser an invisible break opportunity.
+fn insert_zero_width_space_after_git_marker(version: &str) -> String {
+    Regex::new(r"\b(git\w+)([^-\w])(\w+)")
```
colincasey, line 93: minor - you could move this into a one-time compiled regex using `LazyLock`. e.g.; 

```rust
static GIT_MARKER: LazyLock<Regex> = LazyLock::new(|| {
  Regex::new(r"\b(git\w+)([^-\w])(\w+)")
    .expect("regex should compile"); 
  });
```

stack-package-report/src/arch_normalize.rs
```diff
@@ -0,0 +1,38 @@
+//! Collapses architecture-specific package name infixes
+//! (e.g. `binutils-x86-64-linux-gnu` → `binutils-${arch}-linux-gnu`).
+
+const ARCH_PLACEHOLDER: &str = "-${arch}-";
+const ARCH_AARCH64_INFIX: &str = "-aarch64-";
+const ARCH_AMD64_INFIX: &str = "-x86-64-";
```
colincasey, line 6: could these be attached to the `ArchitectureEnum`?

colincasey (changes requested):
are you planning on introducing any integration tests in a follow up PR?

stack-package-report/src/main.rs
Malax: Yeah, that's the exact reason - I wanted this tool to be as lightweight as possible without extra dependencies. But I do think this is a very close tradeoff call here. Maybe we can get another opinion in to tip the scale? @runesoerensen or @edmorley maybe?

stack-package-report/src/stack_package_list.rs
Malax: Yes, that is possible in theory and handled by this tool. Cheverny didn't handle this case gracefully and just plastered over the issue by picking the first version it find for either architecture. However, this should not happen(tm) in practice.

stack-package-report/src/stack_package_list_id.rs
Malax: Yeah, it is not. I implemented it at some point and kept it around to show that the conversion is in theory lossless. But a point can be made to remove it. I'm on the edge about it.

stack-package-report/Cargo.toml
Malax: This was a stylistic choice. We have a couple of lines that would trigger this lint when writing the Markdown contents. It annoyed me, that we used `push_str` in that file at lot and in between we have some `_ = write!()` calls. Since this path of code is not hot at all, I made the tradeoff to get subjectively nicer code for some allocations.

As this lint is not part of the standard lints, but comes from `pedantic`, I felt fine disabling it. When we started to adopt `pedantic`, we said it's fine to disable any lint that comes from that set if necessary.

All that said, happy to change this - it's a highly subjective thing. :)

stack-package-report/src/package_list_devcenter.rs
Malax: That's what I had before, but in my final self-review I chose to remove it to less visual noise. This is not a hot code path, so I traded some extra regex compiles for less visual noise and/or complexity.

stack-package-report/src/package_list_devcenter.rs
Malax: I'm not sure what you're asking, as this test is already testing the format and formatting is the same. What it does not test is the preamble content before the table as that will be prepended by a GitHub action and is static content.

stack-package-report/src/package_list_devcenter.rs
Malax: Not sure what `insta` would get us here for this one test. Is it that the expected string is in a file instead of in code and a prettier diff? We probably don't need filtering of the output, or any other more advanced feature of insta. Since this is just one test, I feel that adding `insta` as a dependency is a bit heavy for what we get from it.

stack-package-report/src/arch_normalize.rs
Malax: Technically of course, but semantically they are _highly_ specific to what this code does and ubuntu package naming conventions. Adding that specific value to a generic `Architecture` enum seems wrong to me.

Malax:
Thanks for the review @colincasey, appreciated! :)

> are you planning on introducing any integration tests in a follow up PR?

I think this is pretty well tested for what it does, what are you looking for in addition? The orchestration logic for adding these artifacts to releases will live in GHA and is a separate PR.

stack-package-report/src/stack_package_list.rs
colincasey: it might be worth validating this assumption somewhere in a test since there are places where you turn lists of packages into maps keyed by package name, right? just to prove you don't accidentally copy what Cheverny does and your code errors (or something) in the rare instance this situation does occur.

stack-package-report/src/stack_package_list_id.rs
colincasey: > kept it around to show that the conversion is in theory lossless

a theory is nice but if you're not gonna need it... 😅

stack-package-report/src/package_list_devcenter.rs
colincasey: i'm wondering if you think the subset you're using is sufficient to prove this is a 1-to-1 replacement? e.g.; you've got two stacks, two architectures, and a runtime/base image set of packages of:
```sh
("apt", "123"),
("binutils-aarch64-linux-gnu", "2.38-4ubuntu2.6"),
("cpp-aarch64-linux-gnu", "4:13.2.0-7ubuntu1"),
("dash", "0.5.11+git20210903+057cd650a4ed-3build1"),
("gcc-13-aarch64-linux-gnu", "13.2.0-23ubuntu4"),
("sudo", "456"),
("sudo-dev", "456"),
("syslinux", "3:6.04~git20190206.bf6db5b4+dfsg1-3ubuntu1"),
```

but if you look at https://devcenter.heroku.com/articles/stack-packages it's obviously a much richer set of data. maybe this is something better suited to a functional test but the point i'm getting at is - would a more comprehensive test that proves this tool produces the exact same content be valuable? i'm sure it's something our robot helpers could churn out pretty easily.

stack-package-report/src/package_list_devcenter.rs
colincasey: i believe you've got other tests that also validate output formatting but agree to disagree 😄

stack-package-report/src/arch_normalize.rs
colincasey: i suppose but why this stood out to me was - if we, for some reason, adding a new architecture, it's not clear that `normalize_arch_in_package_name` would have to also change to account for that.  it looks like you run this over a mix of package architectures+names though so it's not like you could add the target architecture as a parameter since you can't target a specific architecture to normalize for when this is called.

colincasey (approved):
> I think this is pretty well tested for what it does, what are you looking for in addition? 

I trust you to use your judgement @Malax. I'm just pointing out things that jumped out at me like no functional tests.

stack-package-report/src/main.rs
edmorley: Since there is only a single CLI arg, the current implementation seems fine to me - just since every dependency is yet another thing we'll have Dependabot updates for, supply chain risk, compile time hit etc. 

If we ever add any more CLI args handling here apart from this single arg, then yeah lets start using Clap :-)

stack-package-report/Cargo.toml
edmorley: Agree fine to disable this lint. Maybe worth a short 1-line comment here that it's being disabled for style since the code isn't performance sensitive?

stack-package-report/src/package_list_devcenter.rs
```diff
@@ -0,0 +1,523 @@
+//! HTML-table rendering of the current package set across stacks for the
+//! Heroku DevCenter stack-packages article.
```
edmorley:
```suggestion
//! Heroku Dev Center stack-packages article.
```

edmorley (approved):

PR #400, opened 2026-08-12
Explicitly install `libheif-plugin-libde265` on Heroku-26

edmorley:
A customer reported that `libheif-plugin-libde265` was available on Heroku-24 but is missing on Heroku-26, breaking HEIC/HEVC image decoding. This is the libde265 decode plugin for libheif, which backs HEIC support in libvips (ruby-vips / Active Storage, sharp, pyvips), ImageMagick and libgd. Without it these can still decode AVIF (via the aomdec plugin, which remains installed) but not HEIC.

On Heroku-24 the plugin came in as a `Depends` of libheif1, and was made explicit in #388 once libheif1 demoted it to a `Suggests`. When porting to Heroku-26 we deliberately dropped that explicit entry, as we do on a new stack for re-added packages that no longer appear to be used — but this report shows it is still needed. Ubuntu 26.04 ships it in `universe` (enabled on Heroku-26), so re-adding it restores parity with Heroku-24.

Only the runtime plugin is added. libde265-dev is not re-added to the build image: no language package builds against it (they compile against libheif-dev), and resolute's libheif-dev no longer pulls it. The generated lists gain only libde265-0 and libheif-plugin-libde265.

GUS-W-23832742.


runesoerensen (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/3797

